PR #16, opened 2015-06-15
Fix conflict tests

dgenr8:
Double-spends are tracked in the wallet so they can be flagged persistently.  This requires them to be distinguished from malleability clones, but they must not be "swept under the rug" as clones are (clones' accounting metadata is copied from the first clone).

txn_doublespend.py is corrected not to check for the (broken) double-spend sync behavior, and a new test txn_clone.py is introduced which does expect the sync behavior for a true clone.

This change exists as bitcoin/bitcoin#5881, but is needed now in XT.


mikehearn:
@gavinandresen ?

Is there something specific holding up the merge into Core? I see that it's been ignored for over a month now.

dgenr8:
This was against 0.10.2A.  Subsumed by the 0.11 patches.

PR #19, opened 2015-07-07
Blocksize fork

Raystonn:
Can we get this pulled in for an official release?  Gavin's work is solid here.


mikehearn:
Yes, I know, I've been working on the code too. Don't worry, I haven't forgotten ;) There are some bugs that need to be fixed first though. Also, I will write a change to set soft limit=hard limit, as otherwise XT nodes would still create 750kb blocks by default!

PR #20, opened 2015-07-22
Download Tor IPs dynamically

mikehearn:
Default refresh period is 24 hours, controllable via the command line. New dependency on libcurl is added.

This improves robustness against Tor based jamming attacks. I have implemented this attack and tested it against the existing XT code: this version just ensures the data is fresh.


gavinandresen:
I had some time to test today:

Configured and compiled and ran no issues in my OSX development tree.
Running under the debugger, I verified successful download on startup.

BUT: also saw download happening when running -regtest. I haven't thought hard about what modes should disable this, but should definitely not download with -regtest. Should probably also not download if running -tor (or whatever the option is to try to run 'stealth'-- unless libcurl can fetch through a Tor circuit).

Also: extra commits here?

mikehearn:
Yeah I thought about the -tor switch earlier today but haven't fixed it yet. Also should be a switch to disable completely in case of bugs / people who really love the potential for their node getting DoSd.

You can also verify download issues by using -curl-verbose

I'll disable it in regtest too. Originally I left it because differences should ideally be minimised, but I remember now that the regtest suite shuts down and restarts nodes for every test. Plus perhaps in future XT will be in a CI system. So then it can end up being run too frequently.

Thanks for the review!

gavinandresen:
Might be a good idea to delay the fetch at startup a few minutes; I'm thinking of setups that restart the daemon on crash, and what might happen if disk corruption (or something) makes a crash happen at startup.

mikehearn:
The problem is that makes the DoS attack easier (I've prototyped it). It's easy for an attacker who has enough resources to keep connections alive to notice when a node disconnects and then schedule a reconnect in, say, 10 seconds. Then the only way for the attacker to be labelled as low priority is the static IP data, which of course degrades with time.

The ideal fix would be to allow arbitrary ip ranges+priorities to be specified in local files. Then the download thread can save its results to disk and the data is still available after restart. That can be the next improvement. I've been meaning to support disk loading of ranges since the start, but this is another good reason to do so.

mikehearn:
OK, I folded this into the original ipgroups patch and it's now on master. I also added code to disable it in the following situations:

* User specifies -disableipprio
* User has disabled listening via -nolisten
* User has a proxy enabled, which implies you are probably not accepting connections and covers tor/onion mode
* Regtest mode

Methylfenidaat:
Even if it doesn't block anything like Mike says, source code should not contain lists of IP addresses and  or certificates. Also not download lists of IP's to be 'blocked' / 'blacklisted', and not being active by default. 

There are other solutions to protect against DoS, also this does not protect against botnets not using Tor (witch are far cheaper to run compared to slower Tor network). 

Why not make a separate/dedicated app for this? Being hard coded in BitcoinXT it goes against a total open and free network.

christophebiocca:
The attack under discussion is simply "connect repeatedly, from a different IP each time. until you exhaust the connections on the machine". It can be done with just *one* computer hiding behind tor, which *is* cheaper than running a botnet or having multiple IP addresses of your own.

You would have known all this had you scrolled to the bottom of: https://bitcoinxt.software/patches.html

It's right there in the open.

If you have a solution that makes this code unnecessary, submit a pull request.

PR #22, opened 2015-07-27
Implement BIP 101 and accurate sigop/sighash counting

gavinandresen:
Patches against the final 0.11 Core release that implement BIP 101 and implement accurate sigop/sighash counting (as described in draft BIP that hasn't been assigned a number yet).


mikehearn:
Did any of the earlier patches change too, or can I assume my old reviews are still valid for them?

gavinandresen:
Earlier patches are exactly the same, just rebased against the final Core 0.11 release.

gavinandresen:
Updated last commit with better comments in response to @mikehearn review.

src/consensus/params.h
```diff
@@ -25,6 +25,77 @@ struct Params {
...
+    // measure. Prior to the maximum block size fork it was done in two different, ad-hoc,
+    // inaccurate ways.
+    // Post-fork it is done in an accurate way, counting how many ECDSA verify operations
+    // and how many bytes must be hashed to compute signature hashes to validate a block.
+
+    /** Pre-fork consensus rules use an inaccurate method of counting sigops **/
+    uint64_t MaxBlockLegacySigops(uint64_t nBlockTimestamp, uint64_t nSizeForkActivationTime) const {
+        if (nBlockTimestamp < nEarliestSizeForkTime || nBlockTimestamp < nSizeForkActivationTime)
+            return MaxBlockSize(nBlockTimestamp, nSizeForkActivationTime)/50;
+        return std::numeric_limits<uint64_t>::max(); // Post-fork uses accurate method
```
dgenr8: This and further std:: references in this file do not compile for me without adding
```c++
#include <limits>
```
I wonder why that's not necessary for you.

mikehearn: Not for the first time, something doesn't build for you when it does for others. I wonder if the build system is non-deterministic in some way.

mikehearn:
My thinking around the check limit was, what if someone mines an invalid block with an overlimit transaction. In this case there will be a check against the limits every 8 megabytes of data serialized and hashed. If we end up hashing 8MB too much, then it seems like no big deal. As block size goes up the "slack" by which we might verify too much data goes up though, and then the limit can't be made really hard without another consensus rule change. We could end up hashing 16MB too much data (or more, if done in parallel).

However I guess a lot of the overhead is actually not the hashing but rather the serialization. So serializing, then checking the size of the serialised data against the limits, then hashing that, maybe it doesn't save enough time to be worthwhile.

I'll merge this into the only-bigblocks branch and master, and kick off some builds.

src/consensus/params.h
gavinandresen: Hmm.  It should include limits; it works for me because uint256.h includes string, which indirectly, eventually, includes limits.

mikehearn: Tom, what platform / STL are you building with? Perhaps on some STLs <string> does not include <limits>. We should probably spell it out specifically rather than rely on implementation details of the STL headers.

mikehearn:
This was merged and launched already.

PR #23, opened 2015-07-30
Performance fix for IsSuperMajority check

gavinandresen:
The IsSuperMajority check was being done for blocks with ANY of
the SIZE_FORK_VERSION bits set, instead of blocks with ALL of those
bits set.

Thanks to Michael Ruddy for finding during code review.


mikehearn:
I folded this into the main patch. Thanks guys!

PR #24, opened 2015-08-11
Update broken link in readme

alexwaters:
http://xtnodes.com/ is good for non-technical users, also https://bitcoinxt.software/ is showing a 404


mikehearn:
It will stop showing a 404 soon.

PR #31, opened 2015-08-16
MIN_PEER_PROTO_VERSION and p2p network protocol version related code cleanup

mruddy:
MIN_PEER_PROTO_VERSION and p2p network protocol version related code cleanup as discussed with Gavin at https://groups.google.com/forum/#!topic/bitcoin-xt/4z8ojdVonKI

These changes are for a release post-network upgrade to protocol version 70010. Just putting these out there to get it off my todo list.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4262,8 +4262,6 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             return false;
         }
 
-        if (pfrom->nVersion == 10300)
-            pfrom->nVersion = 300;
         if (!vRecv.empty())
             vRecv >> addrFrom >> nNonce;
         if (!vRecv.empty()) {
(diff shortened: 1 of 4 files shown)


mruddy:
I was just reviewing my commit and realized that it's not complete yet.

The reason it's not complete has to do with the serialization of the nTime field in CAddress being treated differently on/after version 31402 and my changing of INIT_PROTO_VERSION from 209 to 70010. That's not backward compatible because the updated XT nodes would expect another four bytes while deserializing the two CAddress instances of the version message. Upgrading INIT_PROTO_VERSION would need another P2P protocol version bump (>70010) and require upgrading peers.

Setting INIT_PROTO_VERSION back to 209 should suffice. I'll get to that later.

---

But, the current handling of the CAddress de/serialization based on protocol version, and then pretending to be an early version for the initial version message of the P2P protocol is awkward. In my opinion, the version message payload after the nVersion should be parsed according to the protocol version that the peer said it was using to communicate as long as the local peer is ok using that protocol.

Basically, that means in the version message processing, only deserialize most of the payload after the MIN_PEER_PROTO_VERSION check, and then also upgrade the send and receive versions before pushing out the "version" and "verack" messages. That is, do this just after deserializing addrMe:
pfrom->ssSend.SetVersion(min(pfrom->nVersion, PROTOCOL_VERSION));
pfrom->SetRecvVersion(min(pfrom->nVersion, PROTOCOL_VERSION));

Also, it seems like the "verack" message is not of much use. The "version" message seems like enough on its own to settle on a protocol version because each node will either get a "version" message or a disconnect from the other node.

I guess anything like this would take a lot of time and have to come down from Core though, right?

PR #32, opened 2015-08-16
tighten up message size sanity checking

mruddy:
tighten up message size sanity checking. add version, filterclear, and reject to maxMessageSizes.

i brought this up a couple of weeks ago at https://github.com/bitcoin/bitcoin/pull/6261#issuecomment-127055875 and didn't get any response so i figured i'd bring it up here (where gavin's patch already made it in).

i got the reject message sizing by looking at the code and referencing:
https://github.com/bitcoin/bips/blob/master/bip-0061.mediawiki
https://bitcoin.org/en/developer-reference#reject

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -58,6 +58,7 @@ BITCOIN_TESTS =\
   test/mruset_tests.cpp \
   test/multisig_tests.cpp \
   test/netbase_tests.cpp \
+  test/p2p_protocol_tests.cpp \
   test/pmt_tests.cpp \
   test/policyestimator_tests.cpp \
   test/pow_tests.cpp \
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3952,11 +3952,16 @@ string GetWarnings(string strFor)
 //
 
 static std::map<std::string, size_t> maxMessageSizes = boost::assign::map_list_of
-    ("getaddr",0)
-    ("mempool",0)
-    ("ping",8)
-    ("pong",8)
+    // values list the max size of each part of the message payload currently defined/used.
+    // values equate to the max payload size for that respective message type.
+    ("getaddr", 0)
+    ("mempool", 0)
+    ("ping", 8)
+    ("pong", 8)
     ("verack", 0)
+    ("version", 4 + 8 + 8 + (4 + 8 + 16 + 2) + (4 + 8 + 16 + 2) + 8 + (3 + 256) + 4 + 1)
+    ("filterclear", 0)
+    ("reject", (1 + 12) + 1 + (1 + 111) + 32) // this is loose max because the max valid is actually 151 bytes as of BIP 61. see the p2p_protocol_tests unit tests.
     ;
 
 bool static SanityCheckMessage(CNode* peer, const CNetMessage& msg)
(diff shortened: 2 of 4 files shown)


mruddy:
@zander fair enough, that makes sense. i just pushed up a commit with the sizes broken out and a better comment added.

mikehearn:
Hey, I'm not ignoring this, but I'd rather @gavinandresen reviews it, and we've both been away last week so unable to do much coding work.

mikehearn:
It looks good to me assuming the numbers are in fact correct.

mruddy:
Cool, I should add some test cases for the additional messages (like Gavin did over at https://github.com/bitcoin/bitcoin/pull/6261/files). I'm pretty sure on the numbers for the version message because I came up with them while working on a DoS tool (a stupid simple one to prove something to myself). It might be a day or two before I get to this.

Related to these changes...in main.cpp, regarding the default case for unknown commands (search for "Ignore unknown commands for extensibility"), should we do anything to limit this case? It's open right now for a bandwidth wasting DoS. All one has to do is send 2MB junk commands in a loop. Or is this not interesting enough to tackle right now?

mikehearn:
Unit tests for the extra cases would be great. There's no hurry.

You can always waste the bandwidth of any computer by firing UDP packets at it. I don't think we should worry about that branch unless there's actually no limit at all, in which case the issue would be wasting RAM and not bandwidth.

mruddy:
I added some new unit tests. I found and fixed a mistake in my version message sizing (I hadn't accounted for four bytes in CAddress before).

In order to make the unit tests, I had to make ProcessMessage not static anymore (minor, but makes patch management more difficult).

Also, to allow testing the reject messages, I added a "return false" to the exception case in the reject message processing. Doing this is mostly uninteresting, but it does allow an extra log message get output so it can allow a disk-fill-up-attack. So maybe it needs so "misbehaving" handling or something. I didn't look to far into it because it was tangential to getting the unit tests done.

Anyways, even if these unit test related changes are not desirable, at least they show that the max sizes in maxMessageSizes for the reject and version messages are logical to use.

gavinandresen:
The "reject" handling should definitely have a call to Misbehaving(pfrom->GetId(), 100) before returning false : a fill-up-disk attack is worse than a waste-bandwidth attack.

Other than that, ACK.

mikehearn:
The debug log is trimmed, I thought. I don't know if we should ban IPs just because we might have logged something, when we could instead just .... not log something.

gavinandresen:
The debug log is trimmed on startup.

RE: not logging: changing:
```LogPrintf("%s(%s, %u bytes) FAILED peer=%d\n"....)```
 to
```LogPrint("net", "%s(%s, %u bytes) FAILED peer=%d\n"....)```

... so bad messages are only logged if you're running -debug=net is a fine fix, too.

mikehearn:
Yeah, I guess we should do that instead.

An RPC to modify the command line arguments on the fly would be very useful. We had that in all servers at Google and it saved the day many, many times.

mruddy:
OK, I made the update to use:
LogPrint("net", "%s(%s, %u bytes) FAILED peer=%d\n"...)

I also rebased/squashed down to a single commit.

I looked at making an RPC to modify the debug flag on the fly. It seems doable on the surface. But, I got to wondering about thread safety (locking and synchronization [memory barrier stuff]) and backed off because it seems like more extensive refactoring would be necessary (and that would be a pain to maintain as a patch). The command line gets parsed into some global maps. Some global flag variables get set based on what's in the maps. Then either can be accessed wherever. Maybe if that was all encapsulated into a configuration management object it would be more sane. But that would have to come down from Core, otherwise it would just be a total pain to patch over and over.

mikehearn:
Merged, thanks!

Agreed on your analysis of the flag RPC changes.

PR #35, opened 2015-08-17
stop adjusting the node's perspective of time based on peer input.

mruddy:
Created this pull request as a result of discussion at https://groups.google.com/forum/#!topic/bitcoin-xt/seVFxVhiGFc

This patch is a more intrusive version that removes the most code. I understand that may cause this to be rejected (if only because it is more difficult to maintain). But, here it is so that can be decided.


dgenr8:
One of node's most important resources is its independent time-keeping hardware.  It should not allow remote nodes to influence its knowledge of the time; this is a clear vector for Sybil attacks.

But this change goes further than to stop adjusting our time.  It also eliminates tracking of the adjustment that the network would recommend.

I think the adjustment calculation (and the tests thereof) should be kept, but used only to alert the user to check his clock.

Reminder (if you are inclined to reinstate the calculation): bitcoin/bitcoin#4521

mruddy:
Yes and no. I mean these changes don't eliminate tracking peer provided clock offsets which I think is the useful data for a node operator. The previous adjustment calculation was/is broken. The offsets are still set on the node objects (in memory) and logged with a new "offset" attribute in the "receive version message" that is output at the end of the version command processing.

christophebiocca:
Seems like a warning doesn't need that calculation anyway. You'd be better off saying ">50% of your peers are 50 minutes ahead of you" or somesuch. The exact offset doesn't matter (and if half your neighbors are in the far future, and half in the far past, that might still be relevant).

dgenr8:
The current alert triggers when the median peer offset is >= 70 minutes, and no peer has a time within 5 minutes of, but not identical to, ours.

Without adjustment, the alert threshold needs to be tighter than 70 minutes because the user is the only fix mechanism.  Ideally it could be set from a distribution of observed offsets.  I'll go analyze some log files.

mruddy:
Right, so the way these changes are implemented, we extract that decision making and alerting and leave it to the node operator. Each operator might want a different alert, or at a different threshold, etc... As long as the data is provided, each can make his own decision. The parameters could be specified by command line args etc..., but really such tracking and alerting is only tangential to the purpose of the bitcoin node software. When I worked at a startup, we would throw the log data into Splunk and then chart, alert, analyze, and change thresholds however we wanted. We used to find that having the basic data points available was the important thing.

gavinandresen:
I always thought the "what if NTP servers get hacked" justification for not changing this was silly, so concept ACK.

But better to keep patches small and as contained-as-possible, so how about just modifying GetAdjustedTime to pay attention to a 'trust my local node's time' option, which can default to TRUE?

mruddy:
OK, I reverted the bigger commit and made a simpler commit that changes the default behavior of GetAdjustedTime. I guess the added overhead of the map lookup will not matter. To use the old/original time adjustment, use: -trustsystemclock=0

dgenr8:
ACK.  I like the precedent of taking action on suspicious patterns seen in the wild.

gavinandresen:
What happens if you run this on a system with a clock that is set wrong? Is the error message reasonable? (IS there an error message that users will see?)

dgenr8:
I verified that a console message (or GUI dialog box) appears with the message "**Warning: Please check that your computer's date and time are correct! If your clock is wrong Bitcoin XT will not work properly.**" at startup and any time new time samples are added, when the clock is different from the network median by 70 minutes or more AND nobody's clock is within 5 minutes of ours.

We may want to change the threshold per https://github.com/bitcoinxt/bitcoinxt/pull/35#issuecomment-132901594, and should also go ahead and fix the bug in bitcoin/bitcoin#4521, since it affects display of the warning.

dgenr8:
Summary of log data collected over 1 month through today
```
93.3% of peer offsets are less than 1 minute
97.6% less than 10 minutes
98.0% less than 30 minutes
99.1% less than 70 minutes
```
There is some minor clustering around +/- 60 minutes that suggests a remote timezone problem.  The 70-minute rule avoids a false positive in that case, but since it takes several peers to drag the median off that much, I think we would be very save in setting the alert threshold at 10 minutes.

mruddy:
@dgenr8 Do you like this KnCMiner commit instead?:
https://github.com/KnCMiner/bitcoin/commit/1f4d1977fb52074dd9766c189d42a677b0a3faaf

I favor removing the time adjustment code entirely (my original commit). I understand that can be a pain to maintain as a patch and that's why I put out the latest "simplest" patch to at least disable it by default. I think I'll leave this commit as an alternative to the KnCMiner (or similar) commit on those grounds. I won't be disappointed if another patch is chosen over the one I've presented here. Just let me know, thanks!

mikehearn:
Alright, looks like we have agreement here on the simple version. I'll merge this.

dgenr8:
Awesome.  This release is shaping up to have a significant set of security enhancements.

PR #41, opened 2015-08-19
Update README.md

baero:
the statement is more effective when communicated in a positive way


mikehearn:
I'm not convinced this is any clearer, to be honest.

baero:
alright, however as an expert in communication I would advise you to rephrase the sentence in a different way. Why? If I told you: "Don't think about a yellow elephant", well, I bet you just thought about it. This is because the subconscious doesn't understand the "not" in a sentence, hence if you say that you don't have to download it you may actually make somebody afraid that something will go wrong.

mikehearn:
Well, alright, fair enough then. I guess you're right - it can't hurt and it might help. I'll merge this. Let us know if you spot any other ways to rephrase things better! :)

PR #46, opened 2015-08-23
gitian: added systemd configs to debian installer creation script

stevemyers:
This change is based on discussion here:
https://groups.google.com/forum/#!topic/bitcoin-xt/DamcJG4jteY

And on TODO comments in build.sh.

1. removed unneed qt, .h and .so files from being installed by the package
2. added a default bitcoin.conf file with a randomly generated rpc password
3. added a systemd startup bitcoinxt.service file
4. added postinst script to fix file owners and permissions, enable the startup script and auto start the bitcoinxt daemon
5. also via the postinst script create "bitcoin" system account and group (with login shell disabled) to run the daemon under
6. added an uninstall (prerm) script that automatically stops the daemon and removes executable files (config files are not removed by default)
7. created README.md with basic instructions for creating and testing debian installer

New scripts and configs tested on fresh install (in virtual box vm) of debian-8.1.0-amd64.

### DIFF
--- a/contrib/gitian-debian/build.sh
+++ b/contrib/gitian-debian/build.sh
@@ -4,8 +4,7 @@
 #
 #  Improvement ideas:
 #  - Install the man page from the source repo.
-#  - Install an init script or systemd file or whatever then hip kids use these days, so it starts at boot.
-#  - Wrap in a script that restarts the node if it crashes and/or sends crash reports/core dumps to some issue tracker.
+#  - Wrap in a script that sends crash reports/core dumps to some issue tracker.
 #  - etc ...
 
 ver=0.11.0
@@ -38,9 +37,9 @@ touch var/lib/bitcoinxt/.empty
 mv usr/bin/bitcoind usr/bin/bitcoinxt
 mv usr/bin/bitcoin-cli usr/bin/bitcoinxt-cli
 mv usr/bin/bitcoin-tx usr/bin/bitcoinxt-tx
+mv usr/bin/bitcoin-qt usr/bin/bitcoinxt-qt
 
 # Remove unneeded files 
-rm usr/bin/bitcoin-qt
 rm usr/bin/test_bitcoin
 rm usr/bin/test_bitcoin-qt
 rm usr/include/*
@@ -53,7 +52,7 @@ cat <<EOF >DEBIAN/control
 Package: bitcoinxt
 Architecture: amd64
 Description: Bitcoin XT is a fully verifying Bitcoin node implementation, based on the sources of Bitcoin Core.
-Maintainer: Mike Hearn <hearn@vinumeris.com>
+Maintainer: Steve Myers <steven.myers@gmail.com>
 Version: $realver
 Depends: adduser, ntp
 EOF
--- a/contrib/gitian-debian/build.sh
+++ b/contrib/gitian-debian/build.sh
@@ -22,10 +22,29 @@ tarball=bitcoin-$ver-linux64.tar.gz
 tar xzvf ../$tarball
 mv bitcoin-$ver usr
 
+# copy bitcoinxt.service file to lib/systemd/system directory
+mkdir -p lib/systemd/system 
+cp ../bitcoinxt.service lib/systemd/system
+
+# copy bitcoin.conf file to etc/bitcoinxt 
+mkdir -p etc/bitcoinxt
+cp ../bitcoin.conf etc/bitcoinxt
+
+# create file to force creation of data folder
+mkdir -p var/lib/bitcoinxt
+touch var/lib/bitcoinxt/.empty
+
 # Rename the binaries so we don't conflict with regular Bitcoin
 mv usr/bin/bitcoind usr/bin/bitcoinxt
 mv usr/bin/bitcoin-cli usr/bin/bitcoinxt-cli
 mv usr/bin/bitcoin-tx usr/bin/bitcoinxt-tx
+
+# Remove unneeded files 
+rm usr/bin/bitcoin-qt
+rm usr/bin/test_bitcoin
+rm usr/bin/test_bitcoin-qt
+rm usr/include/*
+rm usr/lib/*
 
 # Set up debian metadata. There are no dependencies beyond libc and other base DSOs as everything is statically linked.
 
@@ -36,6 +55,7 @@ Architecture: amd64
 Description: Bitcoin XT is a fully verifying Bitcoin node implementation, based on the sources of Bitcoin Core.
 Maintainer: Mike Hearn <hearn@vinumeris.com>
 Version: $realver
+Depends: adduser, ntp
 EOF
 
 cat <<EOF >DEBIAN/install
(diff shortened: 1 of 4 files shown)


stevemyers:
Hi @mikehearn let me know if you want any changes or other testing.

mikehearn:
Thanks Steve! I'll try and make time to review this next week, but it'll be a lower priority than some other issues that have come up. If anyone else wants to help out and review too - please beat me to it. 

Review from someone who understands modern Debian better than I do would be most appreciated.

stevemyers:
Mike, I'd like to get some more review too, and I understand it's not a high priority so no worries if gets it in this week or not.  @zander had some questions on the google group, maybe he can take a look?

contrib/gitian-debian/build.sh
```diff
@@ -22,11 +22,30 @@ tarball=bitcoin-$ver-linux64.tar.gz
...
+mkdir -p var/lib/bitcoinxt
+touch var/lib/bitcoinxt/.empty
+
 # Rename the binaries so we don't conflict with regular Bitcoin
 mv usr/bin/bitcoind usr/bin/bitcoinxt
 mv usr/bin/bitcoin-cli usr/bin/bitcoinxt-cli
 mv usr/bin/bitcoin-tx usr/bin/bitcoinxt-tx
 
+# Remove unneeded files 
+rm usr/bin/bitcoin-qt
```
mikehearn: Should we maybe just rename bitcoin-qt instead of removing it? Some users seem to really want it.

contrib/gitian-debian/build.sh
```diff
@@ -36,6 +55,7 @@ Architecture: amd64
 Description: Bitcoin XT is a fully verifying Bitcoin node implementation, based on the sources of Bitcoin Core.
 Maintainer: Mike Hearn <hearn@vinumeris.com>
```
mikehearn: Can you set yourself as the maintainer maybe? If you're OK with that only, of course!

mikehearn:
This looks really good to me. Do all Debian's use systemd? I thought it was very new. What happens if you install it on an older Debian?

schildbach:
For Ubuntu, only the latest version 15.04 uses systemd.

zander:
Debian 8 uses systemd, older ones do not IIRC.

I suggest adding a dependency on 'systemd' to the dpkg which would solve this issue.
If we then get requests for older debians, we could allow a separate package that uses the older init scripts, assuming we find a volunteer, and just ship 2 .deb files for people to choose.

mikehearn:
Could it maybe not depend on systemd, but probe in the scripts to see if the relevant control program exists? Then it'd work for everyone.

zander:
Technically, yes, you could do this automatically. But the files that are specific to systemd are still installed and testing would become much much harder. And why? Adding the dependency is just making it clear what it already does. It means that people that need a non-systemd install will complain and then we can take action.

Keep it simple ;)

mikehearn:
They'd be installed and ignored until there's an upgrade to the latest version, then they'd activate, right? That sounds about right.

Mainly I just don't want to have to build and upload two packages ....

zander:
I'd expect differences in library versions to cause you to do a build more than once anyway, again, if we actually care about the older versions. My point really is that if you don't know if people out there want you to support an older version of debian, so why put time into that now?

stevemyers:
I did try manually installing systemd and the bitcoinxt service on an older debian system (my arm beagle bone with debian v7), which had been using init.d scripts.  To get it to switch to using systemd I had to install systemd _and_ the systemd-sysv package.  The systemd-sysv package changes the grub boot config to boot with systemd, after installing it and rebooting the systemd bitcoinxt service worked fine.  

I'm reluctant make systemd-sysv a dependency though since it is changes the boot configs and that can cause problems for anyone using custom configs or kernels.  Maybe we can just provide instructions for upgrading a pre-debian v8 systems to use systemd? 

```
$ sudo apt-get install systemd
$ sudo apt-get install systemd-sysv
$ reboot
```

More instructions on the [Debian systemd page](https://wiki.debian.org/systemd#Installing_and_Testing).

General [adoption trend](https://en.wikipedia.org/wiki/Systemd#Adoption_and_reception) seems to be towards systemd.

contrib/gitian-debian/build.sh
stevemyers: Sure I'll add it back in as "bitcoinxt-qt", just need to do a quick test first.

contrib/gitian-debian/build.sh
stevemyers: Sure, no problem.  I'll update the TODO list also to remove what's already completed.

mikehearn:
No need for an init.d script, I was just thinking the pre-inst/post-inst scripts would fail if they tried to run a non-existent command. It's fine for the package to just not configure autostart on older Debians. I mean, the package is great as is, as long as it doesn't puke and die on older distros :)

I didn't realise Debian <8 could be switched to use systemd so easily. That's pretty cool.

@zander The binaries are fully statically linked, so library versions don't make any difference. It's a part of doing things with gitian: the binaries are fully reproducible this way.

stevemyers:
I see what you mean Mike that I need to make sure the pre/post install scripts check for the presents of /bin/systemctl.  I'll put that in and test on a fresh Deb7 vm.  Then installing the package at least won't fail on an init.d based system, it just won't autostart.

mikehearn:
Perfect!

stevemyers:
I've got the built package tested and working now on deb7 and deb8.  On deb7 it skips the systemctl commands and prints a message that the bitcoinxt service was not automatically started.  I also found a weird broken pipe error generating the random rpcpassword which only showed up in deb7.  I fixed that too using the rpcpassword generation method recommended in [doc/build-osx.md](https://github.com/bitcoinxt/bitcoinxt/blob/master/doc/build-osx.md).

mikehearn:
Tested on Debian 7, worked fine. I guess the new ntpd dependency is to ensure the local clock is correct: that seems OK.

Excellent work, thanks a lot! This is now merged.

stevemyers:
Cool thanks Mike!

redfacedquark:
SystemD is the devil. 

Linux sysadmin and developer here trying to help you guys out by running a node. Installed via source, fine except ran out of space as the EC2 instance of the Jessie AMI didn't resize its larger partition on first boot. Noticed the deb second time round having made a second partition ready for data. 

I install the deb and the damn thing starts right away and I cannot stop the service. I hadn't mounted as a non-root user. I kill the process and it starts again. Look at this:

root@ip-172-31-31-11:/home/admin# systemctl |grep bitc
home-admin-.bitcoin.mount                        loaded active mounted   /home/admin/.bitcoin
bitcoinxtd.service                               loaded active running   Bitcoin XT distributed currency daemon
root@ip-172-31-31-11:/home/admin# systemctl stop bitcoinxt.service 
Failed to stop bitcoinxt.service: Unit bitcoinxt.service not loaded.

Going back to source until systemd is less broken.

zander:
noticed a typo in your comment; you asked bitcoin to be stopped, not bitcoind (added d for daemon)

mikehearn:
The service was renamed a couple of times, try 

    systemctl stop bitcoinxtd

Instead of "bitcoinxt" as the service name.

SystemD is what Debian/Ubuntu have switched to, here is not the place for flaming init systems.

stevemyers:
As an FYI for anyone else confused by the naming, I called the deb package bitcoinxt because it installs bitcoinxt-tx, bitcoinxt-cli, bitcoinxt-qt, and the bitcoinxtd service.  The systemd bitcoinxtd.service only manages bitcoinxtd thus its name.

redfacedquark:
Fine, I've been stuck on centos for my main client the last year so if it's just that I didn't know to do ```systemctl list-units|grep bitcoin``` and figure it out fair enough, it's my bad. Not trying to perpetuate a flame war, was just in a rush to get a node up.

Local bitcoind for armory now running XT as well now. Not that it accepts connections.

mikehearn:
list-units is indeed an unintuitive name for seeing what services are available. Ah well. Good to hear you got it sorted out.

PR #49, opened 2015-08-24
Avoid exceeding maximum safe Javascript integer

gavinandresen:
Some JSON libraries cannot parse integers greater than 2^53-1.

So clamp "no limit" 2^64 sigop/sighash limits to 2^53-1 before
returning them.

A couple of implementation notes:

Changing the constant returned by the consensus/params.h methods
from uint64_t::max to 2^53-1 would be another fix, but that felt
like a layering violation: the consensus code shouldn't care about
details of representing numbers in JSON.

Also, I didn't define JSON_NUMBER_LIMIT in a .h file because it
isn't clear where it belongs (util.h? one of the json/ .h?) and
because at this point it makes sense to try to create patches
that change fewer files.

Fixes issue# 45.

After review, should be pulled into the only_bigblocks branch, also.

### DIFF
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -582,10 +582,17 @@ Value getblocktemplate(const Array& params, bool fHelp)
     result.push_back(Pair("mintime", (int64_t)pindexPrev->GetMedianTimePast()+1));
     result.push_back(Pair("mutable", aMutable));
     result.push_back(Pair("noncerange", "00000000ffffffff"));
-    result.push_back(Pair("sigoplimit", Params().GetConsensus().MaxBlockLegacySigops(nBlockTime, sizeForkTime.load())));
     result.push_back(Pair("sizelimit", Params().GetConsensus().MaxBlockSize(nBlockTime, sizeForkTime.load())));
-    result.push_back(Pair("accuratesigoplimit", Params().GetConsensus().MaxBlockAccurateSigops(nBlockTime, sizeForkTime.load())));
-    result.push_back(Pair("sighashlimit", Params().GetConsensus().MaxBlockSighashBytes(nBlockTime, sizeForkTime.load())));
+    // Some JSON libraries can't parse full-64-bit integers. So limit the limits to
+    // the safe javascript number limit, which is 2^53-1
+    // (see http://www.ecma-international.org/ecma-262/6.0/#sec-number.max_safe_integer )
+    uint64_t JSON_NUMBER_LIMIT = 9007199254740991;
+    uint64_t sigoplimit = Params().GetConsensus().MaxBlockLegacySigops(nBlockTime, sizeForkTime.load());
+    result.push_back(Pair("sigoplimit", std::min(JSON_NUMBER_LIMIT, sigoplimit)));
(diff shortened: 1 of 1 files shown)


mikehearn:
LGTM

I asked the bug reporter to test it out. The fix seems like a no-brainer but it'd be nice to ensure there aren't other weird compatibility issues there.

In future it'd be nice to offer something better than JSON for the RPC layer e.g. protobufs. JSON just seems riddled with obscure quirks.

dgenr8:
utACK.  I see, "sigoplimit" would have been a problem post-fork.

gavinandresen:
Tested by running two -regtest instances, having one mine a block (so getblocktemplate doesn't complain about being in initial-block-download state), and then observing getblocktemplate output:
```
gavin$ ./bitcoin-cli -regtest -datadir=$HOME/src/bitcoin/src/.rt2 getblocktemplate
{
    "capabilities" : [
        "proposal"
    ],
    "version" : 536870919,
    "previousblockhash" : "2f7d4ea63ebfd6eec89b3985da6c06cd715b89abc28a79451c8e624148afc9f8",
    "transactions" : [
    ],
    "coinbaseaux" : {
        "flags" : ""
    },
    "coinbasevalue" : 5000000000,
    "longpollid" : "2f7d4ea63ebfd6eec89b3985da6c06cd715b89abc28a79451c8e624148afc9f82",
    "target" : "7fffff0000000000000000000000000000000000000000000000000000000000",
    "mintime" : 1437597387,
    "mutable" : [
        "time",
        "transactions",
        "prevblock"
    ],
    "noncerange" : "00000000ffffffff",
    "sizelimit" : 1000000,
    "sigoplimit" : 20000,
    "accuratesigoplimit" : 9007199254740991,
    "sighashlimit" : 9007199254740991,
    "curtime" : 1440429106,
    "bits" : "207fffff",
    "height" : 4
}
```

Matoking:
Didn't work here. The application throws the following warning during compilation:
```
rpcmining.cpp:592:39: warning: left shift count >= width of type
     uint64_t JSON_NUMBER_LIMIT = (1L<<53)-1;
```

The numbers were still reported as the largest unsigned 64-bit integer instead of 2^53.

Thankfully, the fix was as easy as changing the (1L<<53)-1 to (1L**L**<<53)-1 (long long instead of long).

After that the max values are reported correctly as 9007199254740991. cgminer also connects to the node without any issues after the fix.

gavinandresen:
Replaced the shift-and-subtract with explicit 9007199254740991 constant, which should be the most portable way of doing it (see http://en.cppreference.com/w/cpp/language/integer_literal -- LL suffix is C++11).

Matoking:
Tried with a constant value and it works correctly.

mikehearn:
Perfect, thanks guys.

luke-jr:
I wonder if null would be more appropriate here? Or a string "Infinity" or the boolean false (ie, no limit)?

gavinandresen:
@luke-jr : if you are working in a strongly-typed language, RPC values that can be more than one type or can be missing entirely are painful to deal with. And if they suddenly switch from an integer to a string (as would happen when the fork activates) that's a very good place for a very bad bug to happen if the code interpreting the RPC value isn't written properly.

luke-jr:
Fair enough; I can't think of any practical harm to representing a practically-unbounded limit like this.

PR #50, opened 2015-08-25
Track recently rejected transactions; don't reprocess

dgenr8:
A live DoS attack observed by several nodes in recent days involved repeated rejection of duplicate transactions.

The first six commits to address this problem have already been merged in core bitcoin/bitcoin#6498.  A rolling bloom filter tracks recently rejected transactions to avoid repeatedly requesting them from peers and re-checking them.

The last commit adds a call to AlreadyHave when an *unsolicited* full tx is received, as was the case in the observed attack.  AlreadyHave uses the recentRejects filter.

I tested all of these changes on mainnet while the actual attack was still underway.

The recentRejects filter is cleared when the tip is updated, so nothing stops attacker from re-transmitting a load of rejectable txes after a new block, and in fact our attacker was doing this.  But the duplicates are stopped between blocks and the attack could get arbitrarily heavy if multiple attacking peers were involved.

### DIFF
--- a/src/bloom.cpp
+++ b/src/bloom.cpp
@@ -236,16 +236,8 @@ void CRollingBloomFilter::insert(const std::vector<unsigned char>& vKey)
 
 void CRollingBloomFilter::insert(const uint256& hash)
 {
-    if (nInsertions == 0) {
-        b1.clear();
-    } else if (nInsertions == nBloomSize / 2) {
-        b2.clear();
-    }
-    b1.insert(hash);
-    b2.insert(hash);
-    if (++nInsertions == nBloomSize) {
-        nInsertions = 0;
-    }
+    vector<unsigned char> data(hash.begin(), hash.end());
+    insert(data);
 }
 
 bool CRollingBloomFilter::contains(const std::vector<unsigned char>& vKey) const
@@ -258,10 +250,8 @@ bool CRollingBloomFilter::contains(const std::vector<unsigned char>& vKey) const
 
 bool CRollingBloomFilter::contains(const uint256& hash) const
 {
-    if (nInsertions < nBloomSize / 2) {
-        return b2.contains(hash);
-    }
-    return b1.contains(hash);
+    vector<unsigned char> data(hash.begin(), hash.end());
+    return contains(data);
 }
 
 void CRollingBloomFilter::clear()
--- a/src/bloom.cpp
+++ b/src/bloom.cpp
@@ -216,16 +216,17 @@ void CBloomFilter::UpdateEmptyFull()
     isEmpty = empty;
 }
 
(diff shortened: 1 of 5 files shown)


dgenr8:
Last commit submitted to core as bitcoin/bitcoin#6588.

mikehearn:
Does this attack actually deny service in any way? Or is it just a waste of bandwidth?

If it's just a waste of bandwidth, why is this the right fix? The attacker can just switch to sending large UDP packets or some other kind of bandwidth flood.

dgenr8:
Transactions are re-validated every time before they are rejected.  Depending on the rejection reason, a varying amount of local resources are spent.

So far, I have not seen it be successful enough to impact service at all.  However, most of it has been coming from only a single peer...

mikehearn:
If the attack isn't capable of much and it's already upstream, I think for now perhaps we should just wait and pick it up in the next rebase. Thoughts?

dgenr8:
XT nodes were the ones actually being attacked.  a261af5 looks on track to be included in core.  We should get this into 0.11B.

dgenr8:
During today's spam attack, in the first 20 minutes after my 0.11B node mempool filled up, it dropped 187000 low fee / low pri txes (and evicted only 2000 during the same time period).

It's likely that many of the dropped txes were duplicates.  CPU utilitzation was at 45% vs. 15% normally (earlier, while the mempool was still filling up -- it was at 100%).

No service denied.  But since the pool-full check is last, resources were spent, which this change would have avoided.

mikehearn:
CPU usage was at 100% when the mempool was filling? So we ran out of CPU to check inbound transactions completely? Perhaps it's time to multi-thread that code path.

gavinandresen:
The low-level signature checker is multithreaded already, and defaults to
using as many cores as you have.

On Mon, Sep 14, 2015 at 9:16 AM, Mike Hearn <notifications@github.com>
wrote:

> CPU usage was at 100% when the mempool was filling? So we ran out of CPU
> to check inbound transactions completely? Perhaps it's time to multi-thread
> that code path.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitcoinxt/bitcoinxt/pull/50#issuecomment-140071256>.
>



-- 
--
Gavin Andresen

mikehearn:
@gavinandresen I believe it's only multi-threaded when checking blocks. Look at the CheckInputs calls in AcceptToMemoryPool - the pvChecks parameter is NULL and when it's NULL the checks are performed serially inside a for loop over the inputs.

gavinandresen:
Ah, right-- another symptom of the code getting optimized for initial block download, not tx/block validation/propagation....

dgenr8:
The core version of 6815266 has been merged.

PR #52, opened 2015-08-30
Fix init scripts

keo:
fixed init scripts so they work with the new binary name.


dgenr8:
ping @stevemyers

stevemyers:
Good idea @keo, I also suggest if we're going to rename these files we should change the daemon name inside these files also to ```bitcoinxt```.  

And this also brings up another topic I'd like to get feed back from @mikehearn and @gavinandresen, are you happy with the daemon being called ```bitcoinxt```, or should we change it to ```bitcoinxtd``` ?  Ending the filename with a ```d``` is more in line with the typical unix daemon naming convention.   I realize this is a minor quibble but changing file name shouldn't be that hard and improves clarity for its function.

mikehearn:
Don't care either way, xtd is also fine by me. You're the package maintainer - use the power ;)

mikehearn:
Thanks keo!

keo:
thanks for merging - my VM building scripts now work directly from the git repo. :)

stevemyers:
I've updated bitcoinxt to bitcoinxtd in #55, anyone who's interested please review.  Thanks!

PR #54, opened 2015-08-31
IP priority improvements

mikehearn:
* Can now load IP prio data from local files, specified via a new -ip-priority-source command line flag. It's not currently documented as the whole framework is still quite young and the flag will probably get new features in future like the ability to use a URL as well as a file.
* Switch the Tor URL used based on feedback from Isis Lovecruft. This also means we can use a simple unified IP format now where comment lines start with a hash and the rest are just IPs or netblock specifications.
* Tor data is no longer fetched if we don't have any routeable IPs. This means if you're behind a firewall and can't listen anyway, it won't put any load on the Tor servers even if the node itself doesn't have listening disabled or a proxy configured.
* Prio data is now calculated on demand instead of being cached, so reducing memory usage and making the framework more adaptive.
* Added a TODO comment to the network code about handling IPs that make many connections.

There are still many other things that can be improved.

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -58,6 +58,7 @@ BITCOIN_TESTS =\
   test/mruset_tests.cpp \
   test/multisig_tests.cpp \
   test/netbase_tests.cpp \
+  test/p2p_protocol_tests.cpp \
   test/pmt_tests.cpp \
   test/policyestimator_tests.cpp \
   test/pow_tests.cpp \
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -295,6 +295,7 @@ std::string HelpMessage(HelpMessageMode mode)
 #if !defined(WIN32)
     strUsage += HelpMessageOpt("-sysperms", _("Create new files with system default permissions, instead of umask 077 (only effective with disabled wallet functionality)"));
 #endif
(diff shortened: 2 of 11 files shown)


zander:
awesome stuff!

I was just thinking; would we be able to make it so that when a node hits, say, 80% of its incoming connections used that only then it considers downloading the prio list?

mikehearn:
We could yeah, but the ip prio stuff may be useful for more than just being full in future. For instance, let's say someone finds a way to hog a lot of CPU or such and we don't want to solve this by just banning the IP, because that's bad for shared IPs. Then we can also use it to do socket or CPU scheduling as well as RAM scheduling.

I don't want to put too much effort into the downloading framework. As noted already, we need to find more dynamic approaches. For instance, if an idle connection that isn't relaying valid transactions/blocks could drag down its IP's priority over time. This is trickier to get right because mobile / desktop clients also sit around not doing much beyond waiting to see if they receive money, but we kind of like users who are expecting to get paid in Bitcoin - they're good for our economy. And an attacker who wants to DoS the network can also relay stuff from time to time (pure relaying without listening isn't all that useful, it just adds latency).

gavinandresen:
ACK-- tested on a machine behind a firewall, and it did not download anything.

Also did fuzz testing of the IP parsing code.

mikehearn:
Thanks for the review Thomas, and thanks for the fuzz testing Gavin. I've applied the review comments except for switching to an iterator, I think the current code is easier to read. I'll merge the improved code.

PR #56, opened 2015-09-02
Limit number of transactions in the memory pool

gavinandresen:
Implement a simple strategy to keep the memory pool from consuming
all of memory, by limiting the number of tranasctions stored in
the memory pool.

By default, the limit is 6-blocks-worth of average-sized (500-byte)
transactionsl (12,000 transactions for 1MB blocks). Command-line
option -maxmempooltx overrides the default (and is useful for testing).

When the limit is hit, low-priority transactions are dropped
immediately. Higher priority transactions are added, and
a random transaction (and their descendants) are evicted
from the mempool.

BUT, transactions generated or submitted by ourselves (via
sendtoaddress/sendmany/sendrawtransation) are always allowed into
the mempool and are never evicted. And transactions marked as
important via the prioritsetransaction RPC call are also never
evicted.

To test:

Run with -maxmempooltx=500 -debug=mempool
example debug.log output:
```
2015-09-02 13:17:55 mempool full, dropped tx 0185aa01c280e5bf47fc5e586c1beb712bed3a3ce0f3ac57251bbc49ed786d79
2015-09-02 13:17:55 mempool full, dropped tx e852c5f2430357ee7485f927f189218fffb73368715e5b884afb31a10ccdac68
2015-09-02 13:17:55 mempool full, dropped tx 0185aa01c280e5bf47fc5e586c1beb712bed3a3ce0f3ac57251bbc49ed786d79
2015-09-02 13:17:56 mempool full, dropped tx 30badbcc9721e2d78c480c8b81a4de2274daf6a95b7ff3b32ab40d00dbd2070f
2015-09-02 13:17:56 mempool full, evicted tx e388638f6fb34dc072a4526ba2a5cd34a6e8169bb3d75c083a16d0990667921d
2015-09-02 13:17:56 AcceptToMemoryPool: peer=3 /Satoshi:0.11.0/: accepted 7a8bee60593727b4e74033ace208fa463400d7d269b514b7ff2beca31890957b (poolsz 500)
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1004,7 +1004,9 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
     if (pfMissingInputs)
         *pfMissingInputs = false;
 
-    if (!CheckTransaction(tx, state, Params().GetConsensus().MaxBlockSize(GetAdjustedTime(), sizeForkTime.load())))
+    int64_t maxBlockSize = Params().GetConsensus().MaxBlockSize(GetAdjustedTime(), sizeForkTime.load());
+
+    if (!CheckTransaction(tx, state, maxBlockSize))
         return error("AcceptToMemoryPool: CheckTransaction failed");
 
     // Coinbase is only valid in a block, not as a loose transaction
@@ -1183,8 +1185,51 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         }
         else
         {
-            // Store transaction in memory
-            pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+            // Default max mempool #tx: 25-blocks-worth of 500-byte transactions:
+            int64_t nDefaultMax = 25 * (int64_t)(maxBlockSize / 500);
+            int64_t nMaxPoolTx = GetArg("-maxmempooltx", nDefaultMax);
+
+            if (nMaxPoolTx <= 0) { // Zero or negative: don't limit
+                pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+            }
+            // fLimitFree is false when transactions are submitted from the wallet
+            // code or RPC send commands. We always want to add them to the pool,
+            // and want to prioritise them so they're never evicted:
+            else if (!fLimitFree) {
+                pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+                CAmount unused;
+                pool.PrioritiseTransaction(hash, hash.ToString(), AllowFreeThreshold(), unused);
+            }
+            // Mempool not full:
+            else if ((int64_t)pool.size() < nMaxPoolTx) {
+                pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+            }
+            else { // Mempool full...
+                static int64_t lastEstimate = 0; // Only call estimateFee every 10 minutes
+                static CFeeRate highFee;
+                if (entry.GetTime()-lastEstimate > 600) {
+                    lastEstimate = entry.GetTime();
+                    highFee = pool.estimateFee(2);
+                }
+                if (dPriority < AllowFreeThreshold() && CFeeRate(nFees, nSize) < highFee) {
+                    // Low-priority, low-fee: dropped immediately
+                    LogPrint("mempool", "mempool full, dropped %s\n", hash.ToString());
+                    SyncWithWallets(tx, NULL, false); // Let wallet know it exists
+                    return false;
+                }
+                // High enough priority/fee: add to pool, we'll evict somebody below
+                pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+            }
+            if (nMaxPoolTx > 0 && (int64_t)pool.size() > nMaxPoolTx) {
+                list<CTransaction> evicted;
+                pool.evictRandom(evicted);
+                BOOST_FOREACH(const CTransaction &etx, evicted) {
+                    LogPrint("mempool", "mempool full, evicted %s\n", etx.GetHash().ToString());
+                    SyncWithWallets(etx, NULL, false);
+                }
+                if (!pool.exists(hash))
+                    return false;
+            }
         }
     }
 
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -219,6 +219,24 @@ void CTxMemPool::removeForBlock(const std::vector<CTransaction>& vtx, unsigned i
     minerPolicyEstimator->processBlock(nBlockHeight, entries, fCurrentEstimate);
 }
 
+void CTxMemPool::evictRandom(std::list<CTransaction>& evicted)
+{
+    LOCK(cs);
+    if (mapTx.empty())
+        return;
+    for (int i = 0; i < 11; i++) {
+        map<uint256, CTxMemPoolEntry>::iterator it = mapTx.lower_bound(GetRandHash());
+        if (it == mapTx.end())
+            it = mapTx.begin();
+        if (mapDeltas.find(it->first) != mapDeltas.end())
+            continue;  // prioritised: try again
+        remove(it->second.GetTx(), evicted, true);
+        return;
+    }
+    // There is a small chance nothing will be evicted if the mempool is mostly
+    // prioritised transactions. That's OK.
+}
+
 void CTxMemPool::clear()
 {
     LOCK(cs);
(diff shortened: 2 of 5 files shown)


src/txmempool.h
```diff
@@ -118,6 +118,10 @@ class CTxMemPool
     void removeConflicts(const CTransaction &tx, std::list<CTransaction>& removed);
     void removeForBlock(const std::vector<CTransaction>& vtx, unsigned int nBlockHeight,
                         std::list<CTransaction>& conflicts, bool fCurrentEstimate = true);
+    // evict a random transaction and its descendants, given a random transaction ID,
+    // BUT does not evict prioritised transactions. Returns list of transactions
+    // evicted.
+    void evictRandom(const uint256& hash, std::list<CTransaction>& evicted);
```
mikehearn: Why is the "random transaction ID" here passed as an argument? Wouldn't it make more sense for evictRandom to generate the random key itself?

src/txmempool.cpp
```diff
@@ -219,6 +219,18 @@ void CTxMemPool::removeForBlock(const std::vector<CTransaction>& vtx, unsigned i
...
 
+void CTxMemPool::evictRandom(const uint256& randomhash, std::list<CTransaction>& evicted)
+{
+    LOCK(cs);
+    if (mapTx.empty())
+        return;
+    map<uint256, CTxMemPoolEntry>::iterator it = mapTx.lower_bound(randomhash);
+    if (it == mapTx.end())
+        it = mapTx.begin();
+    if (mapDeltas.find(it->first) == mapDeltas.end())
```
mikehearn: // Avoid evicting transactions that were manually prioritised via RPC.

mikehearn:
For safety, it's nice to be able to disable new features from the command line.

In this case, I assume setting the new flag to some really high value is good enough, although it won't disable if the if(!fLimitFree) codepath. Perhaps we could have the old codepath be invoked if the flag is set to -1 or something, to make disabling easier/more reliable?

mikehearn:
Another thought - 6 feels like a very low default. We routinely see 6 block backlogs just due to natural variance. I'd feel much better if this code only triggered during real attacks, which would imply a much bigger value, perhaps more like 30 or 40ish. Tradeblock is claiming that 28,000 transactions uses about 35mb of space, which isn't a concerning amount of storage. So perhaps more like 50,000 transactions is a more appropriate default?

mikehearn:
Looks good to me.

Will run this+other pending pulls on my server overnight (will crank down the mempool limit to make sure that gets tested too).

zander:
Looks good now. Thanks!

mruddy:
It looks like these changes allow the mempool to become inconsistent with the transaction relay map (specifically looking at the two RelayTransaction functions in net.cpp). It appears that after fifteen minutes, the relay messages will expire. Was this done purposefully to avoid a race between when a peer node receives an "inv" message from us and when we go to respond to that peer's "getdata" (for a transaction that may have just been randomly evicted)?

gavinandresen:
@mruddy : the relay map and the mempool could get out-of-sync before this pull, and yes, that's by design-- the relay map just supports the inv/getdata cycle.

Added a SQUASHME commit that I'll squash tomorrow, assuming the run-overnight test goes well.

The Core commit from @morcos that tackles this same problem is:
  https://github.com/bitcoin/bitcoin/pull/6470

I don't claim that this commit is better in any dimension besides code complexity.

shayanb:
I think this PR would increase the double spending attack surface, specially with the combination of double spend relaying.

mruddy:
@gavinandresen Yep, I was just going to edit my comment to say that "allows" was not the right way to phrase that (but you beat me to it). My concern was about memory pressure from evicted transactions being left in that other map. Thanks for confirming the design though, that makes sense.

gavinandresen:
@shayanb : What attack are you thinking of?

I'm pushing a commit so the wallet is told about low-priority transactions, even if they're immediately dropped.

gavinandresen:
I've saved a debug.log from my node running with -maxmempooltx=2000 during today's spam attack, from right where the mempool starts to overflow:  https://gist.github.com/gavinandresen/0612e1fbbfc9c53a9942

Note that transactions on the end of long chains of unconfirmed transactions are more likely to get evicted (e.g. see the chain starting at txid 035e3ce96 in the gist), and, once part of the chain is evicted from their peers' memory pools, further transactions in the chain get stuck in mapOrphans and are not relayed. This makes it much more difficult for attackers to sustain effective attacks.

mikehearn:
Hmm, many of the transactions rated as low priority seem to have been paying high fees and were subsequently mined. I'd have expected way more random evictions and way fewer immediate drops.

Should that branch also have a fee check on it?

shayanb:
@gavinandresen I think you answered most of my questions in your reply to Moe Adham's question in [bitcoin-xt google groups thread](https://groups.google.com/forum/#!topic/bitcoin-xt/stN6IXM_C0Y).

also I agree with @mikehearn on the fact that some checks on the fee are needed, just in case a similar to low-priority transaction is actually not low-priority and needs to be included in the block, the sender could include higher miner's fee to have higher priority transaction (e.g. miner pool's payouts).

gavinandresen:
Added code to treat high-fee ("should be confirmed in the next block or two" according to the fee estimation code) as high-priority.

Result is many more evictions and many fewer immediate drops; the "how high must the fee be" threshold could be tweaked, but @mikehearn 's mempool explorer isn't working for me

(gradlew run fails with:
```
:compileKotlin
w: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (191, 30): Parameter 'ev' is never used
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (290, 24): Unresolved reference: addListener
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (290, 45): Unresolved reference: AbstractBlockChainListener
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (294, 17): 'receiveFromBlock' overrides nothing
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (309, 17): 'notifyNewBestBlock' overrides nothing
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (328, 27): Unresolved reference: DownloadProgressTracker
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (338, 18): Unresolved reference: addEventListener
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (338, 44): Unresolved reference: AbstractPeerEventListener
e: /Users/gavin/src/Mempool-Explorer/src/main/kotlin/mpexplorer/Main.kt: (339, 17): 'onTransaction' overrides nothing
:compileKotlin FAILED

FAILURE: Build failed with an exception.
```

mikehearn:
Yes sorry, I just noticed that myself and committed some fixes. I had some un-checked in code locally. If you do a "git pull" and rerun "./gradlew run" it should work now, I think.

src/main.cpp
```diff
@@ -1183,8 +1185,49 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
+            }
+            // Mempool not full:
+            else if ((int64_t)pool.size() < nMaxPoolTx) {
+                pool.addUnchecked(hash, entry, !IsInitialBlockDownload());
+            }
+            else { // Mempool full...
+                static int64_t lastEstimate = 0; // Only call estimateFee every 10 minutes
+                static CFeeRate highFee;
+                if (entry.GetTime()-lastEstimate > 600)
+                    highFee = pool.estimateFee(2);
```
dgenr8: Need to update lastEstimate here.

gavinandresen:
Fixed the bug @dgenr8 found that caused estimateFee to be called every transaction.

And squashed all the intermediate commits and rebased to master.

mikehearn:
Merged this, thanks!

dgenr8:
Running live for 15 hours, nothing has been ejected (I have not done positive tests of ejection).

mikehearn:
Great. I've done tests with lowered sizes and seen ejection working properly. I wonder how the 2-block-fee-estimation logic interacts with the full pool in DoS flood scenarios though: I understand the rationale for it, but it seems additional complexity vs pure random rejection. The fee estimator tells you what other people think the right answer is, rather than actually giving the right answer (where right answer==min fee to confirm within time interval).

dgenr8:
High-fee is a check against immediate drop of a low priority tx when pool is full.  Without this, an earnest tx sender has no way to get in.

PR #58, opened 2015-09-05
Survey

dagurval:
This is an opt-in for XT survey that users can enable to poll our website and tell us they are running an XT node. This was suggested on the mailing list (not by me) as a method of knowing how many XT nodes are out there while still in stealth mode.

If _in_ stealth mode, I've made it an opt-out. The defaults can be discussed.

As a bonus, as also suggested on the mailing list, this can be used to notify users that a new version is available.

For now I have not implemented a backend for this (the website). It's only tested with the unittests I've written. I don't have a place to host such backend. I'm willing to work on the backend if this is to-be-merged.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -99,6 +99,7 @@ BITCOIN_CORE_H = \
   hash.h \
   init.h \
   ipgroups.h \
+  curl_wrapper.h \
   key.h \
   keystore.h \
   leveldbwrapper.h \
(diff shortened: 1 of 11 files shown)


mikehearn:
Thanks! Very cool that you've stepped up to work on this.

Let's get the first commit into the right shape and get that in. The survey code is more complicated. I don't think we want a dynamic backend. Just logging user-agents is good enough to start with. Version alerts and the like can come in a release after the next one, as I'd like to do 0.11B really soon.

So do you have time to do the curl_wrapper fixes in the next day or two? If not, don't worry, I'll do them myself.

dagurval:
Thanks for the code review! I'll see if I have the time this evening to fix the issues you pointed out. If not, I'll contact you since it's time sensitive.

I agree that a static back-end is good enough start for survey. Keep it simple!

dagurval:
Opened a new pull request (#60) with just the curl change. Closing this one.

PR #59, opened 2015-09-07
Do one lookup instead of two

zander:
Avoid doing up to 3 map-lookups based on a string (which is not the cheapest) and replace it with doing only exactly one lookup.

Other than that, the logic is unchanged.


mikehearn:
Nice one.

mikehearn:
Why closed? I didn't merge it yet.

zander:
You cherry-picked and pushed it ;)

https://github.com/bitcoinxt/bitcoinxt/commit/b7c08bea08300e4695e09e7848909137776cf271

PR #60, opened 2015-09-07
Extract curlwrapper from ipgroup for re-use

dagurval:
This PR has fixes for issues pointed out by Mike in review. I squashed the fixes into the previous commit, to see just the changes see this commit: https://github.com/dagurval/bitcoinxt/commit/9b5de8d94b139fdbe91dd4d17ec38913bfe30ea3

Changes:
- Handle errors from curl_easy_perform
- Throw errors when status code is not 200 OK
- Expose DummyCurlWrapper to simplify the main interface

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -99,6 +99,7 @@ BITCOIN_CORE_H = \
   hash.h \
   init.h \
   ipgroups.h \
+  curl_wrapper.h \
   key.h \
   keystore.h \
   leveldbwrapper.h \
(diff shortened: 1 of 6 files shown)


mikehearn:
Untested but this looks good to me.

Would like another review.

mikehearn:
@dagurval whose court is the ball in here? Is this waiting for me to take another look or are there still pending changes you'd like to make?

dagurval:
I pushed a commit that I think fixes issues and adds suggestions pointed out by @zander. Correct me if I'm wrong. A exception is not adding a setOpt method. I can't think of a good way to wrap setOpt since curl_easy_setopt is a variadic function. 

If this looks OK, I'll squash it with the previous commit.

dgenr8:
@dagurval Please rebase or close.

dagurval:
It's already in master :+1:

PR #64, opened 2015-09-10
Cleanup splash

zander:
Thu Sep 10 10:02:04 2015 zander <thomas@thomaszander.se>
 ID 4d8e85efe09bf4b04cd6f181020a10c41a821021
  Make the splashscreen not appear jagged anymore

Thu Sep 10 09:51:43 2015 zander <thomas@thomaszander.se>
 ID 429561b0949b91a53896dad1fdd17209634c42d5
  Cleanup the NetworkStyle a little.
  
  The class was created using a scopedPointer so it would be deleted
  when it went out of scope at the end of the method.
  A much easier way to do this is to use an on-stack instance and that
  also means we don't pass a pointer to the constructor of various classes
  having unclear ownership and lifetime.
  
  I also removed many usages of pixmap which is very expensive to use
  for image manipulations.  A pixmap implicitly uploads its data to the
  graphics card memory, so only use it for diplaying purposes!
  
  Last I ran the bitcoin.png through an optimizer which reduced the size
  by about 30%


src/qt/networkstyle.cpp
```diff
@@ -20,78 +23,72 @@ static const struct {
     {"test", QAPP_APP_NAME_TESTNET, 70, 30, QT_TRANSLATE_NOOP("SplashScreen", "[testnet]")},
     {"regtest", QAPP_APP_NAME_TESTNET, 160, 30, "[regtest]"}
 };
-static const unsigned network_styles_count = sizeof(network_styles)/sizeof(*network_styles);
+static const int network_styles_count = sizeof(network_styles)/sizeof(*network_styles);
```
jonasschnelli: `unsigned int`?

src/qt/networkstyle.cpp
```diff
@@ -20,78 +23,72 @@ static const struct {
...
+            iconColorSaturationReduction = network_styles[x].iconColorSaturationReduction;
+            break;
         }
     }
-    return 0;
+    if (appName.isEmpty())
+        throw std::runtime_error("Unknown networkId passed into NetworkStyle");
+
+    appIcon = fixIcon(QImage(":/icons/bitcoin"), iconColorHueShift, iconColorSaturationReduction);
+    Q_ASSERT(appIcon.width() == 1024);
```
jonasschnelli: sure about this? If there a drawback when using a different (larger) image size?

src/qt/networkstyle.cpp
```diff
@@ -20,78 +23,72 @@ static const struct {
...
+    Q_ASSERT(appIcon.height() == 1024);
+    trayAndWindowIcon = QIcon(QPixmap::fromImage(appIcon.scaled(256, 256)));
```
jonasschnelli: `QIcon` could be avoided by changing `void MacDockIconHandler::setIcon(const QIcon &icon)` to `void MacDockIconHandler::setIcon(const QPixmap &pixmap)` (but required a 128x128 icon otherwise it might be invisible [had similar issues recently]).

src/qt/networkstyle.cpp
zander: why unsigned? Its not needed and almost all Qt apis avoid unsigned, so it would be inconsistent.

src/qt/networkstyle.cpp
zander: I didn't add the requirement, it was already there. I just added an assert for it.

The code already assumed 1024, but never tested it.  This assert is here to make sure that the png we put inside the executable is indeed the right size.  If nothing else because we need to review the code if the size changes.  It would break, for instance, if the aspect ration were to change.

src/qt/networkstyle.cpp
zander: We still need QIcon for the other platforms.

For your other point, see the TODO below :)

mikehearn:
This looks good to me. 

@jonasschnelli Thanks for the review, I assume you don't have any pressing objections? It's been a loooooong time since I used Qt so will delegate to your fine judgement.

mikehearn:
OK, as there were no objections, I'll test and review this once my plate is cleared of other higher priority items. We have not forgotten about it!

PR #65, opened 2015-09-10
Allow building out-of-source

zander:
Its quite nice to have a with-wallet and a disable-wallet build tree
that come from the same directory. So lets make sure we can build
out of source (mostly) again.

Seems that levelDB is not able to, so that one will be shared. But
levelDB development should not happen inside of bitcoin anyway, so
thats fine.


mikehearn:
LGTM

mikehearn:
Merged, thanks!

PR #66, opened 2015-09-10
Raise RPC window when opened again.

zander:
If the RCP console is open, but not visible, raise it and activate it.
Before we just seemed to ignore the user request.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2011-2014 The Bitcoin Core developers
+// Copyright (c) 2015 The Bitcoin XT developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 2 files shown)


mikehearn:
Hmm, seems there are lots of ENABLE_WALLET changes in here too?

@jonasschnelli What do you think? Would you mind taking a quick look?

jonasschnelli:
Same points. Maybe keep align with core and just pick https://github.com/bitcoin/bitcoin/pull/6594 (solves the reopening issue)

zander:
The changes in ENABLE_WALLET are based on the messyness of the code, this cleans that part up quite a bit.
The new callback (slot) fits in better now.

jonasschnelli:
Agreed.
But adding a QT code delta on xt, would xt disallow to reuse this changes if the core-wallet gets SPV and gets independent from the node.
I would highly recommend to add such changes to core (even if it's harder to get reviews and merges).

zander:
I didn't intend the code to be for core, sorry.

PR #67, opened 2015-09-10
Fix API issues

zander:
Follow the Qt standard for memory management; pass the QObject
inheriting object as the last argument and default it to zero.

Also remove one place where we use a boolean to an enum, which helps
reading code.

### DIFF
--- a/src/qt/addressbookpage.h
+++ b/src/qt/addressbookpage.h
@@ -39,7 +39,7 @@ public:
         ForEditing  /**< Open address book for editing */
     };
 
-    explicit AddressBookPage(Mode mode, Tabs tab, QWidget *parent);
+    explicit AddressBookPage(Mode mode, Tabs tab, QWidget *parent = 0);
     ~AddressBookPage();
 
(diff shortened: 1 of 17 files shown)


mikehearn:
Again, would like a review from @jonasschnelli as I'm not all that familiar with Qt.

I wonder if these sorts of small UI changes are best done in XT or submitted to Core. As far as I can tell, small code cleanups and refactorings to the UI don't get bogged down in the usual Core patch flamewars and in the end if we have lots of them someone (probably me) will end up having to rebase all these patches to follow changes in the upstream Qt code. That will be potentially a big time sink. It'd be best to avoid forking the UI code given that there's no real disagreements in this region.

jonasschnelli:
What is the value of this PR?

The question is, if you want to do such trivial / refactoring changes and distinct the QT code between xt and core (result maybe in difficult non-automatic merges).
I would recommend to keep the QT/wallet part aligned with core as much as possible.

Codewise: utACK

PR #71, opened 2015-09-13
gitian: ask user if they want to auto-start bitcoinxtd service during deb pkg installation

stevemyers:
1. moved debian installation scripts (```postinit```, ```prerm```) out of ```build.sh``` file to separate files
2. added ```debconf``` as a package dependency
3.  updated ```postinit``` script to ask user if they want to autostart bitcoinxtd and act accordingly 
4. added debconf ```templates``` file with autostart question verbiage
5. added ```postrm``` script to clear debconf data if package removed with "purge" option

Tested on Debian 8 (with systemd) and Debian 7 (no systemd) systems.

Note: Once user selects Yes or No for autostarting bitcoinxtd during installation they will not be prompted again unless they uninstall bitcoinxt with ```sudo apt-get purge bitcoinxt```.

### DIFF
--- a/contrib/gitian-debian/README.md
+++ b/contrib/gitian-debian/README.md
@@ -64,3 +64,17 @@
   sudo apt-install purge bitcoinxt
   sudo rm -rf /var/lib/bitcoinxt
   ```
+
+**Non-Interactive Installation**
+
+The bitcoinxt debian package uses debconf to ask the user if they want to automatically enable and start the bitcoinxtd service as part of the package installation. To skip this question for non-interactive installs the following instructions allow you to pre-answer the question. This question is only asked the first time the bitcoinxt package is installed and only if the target system has the systemd systemctl binary present and executable.
+
+1. Install ```debconf-utils```
+ ```
+ % sudo apt-get install debconf-utils
+ ```
+
+2. Pre-answer the question, ***true*** to automatically enable and start the ```bitcoinxtd``` service and ***false*** to not automatically enable and start the service during package install
+ ```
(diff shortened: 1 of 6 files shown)


Beanow:
Since I'm not familiar with the debconf priorities, have you tested whether the `db_input high ...` results in an acceptable `apt-get install -y bitcoinxt` non-interactive default?

stevemyers:
@Beanow unfortunately the default ```debconf``` behavior is to only ask users questions marked with priority ```high```.  I've tried testing with ```sudo gdebi -o APT::Get::Assume-Yes=true``` to simulate ```apt-get install -y``` , but that option seems to be ignored for ```debconf``` questions.  Since I believe the most common installation case is interactive I'm leaving the ```db_input``` priority ```high```.  

For those more technical users who want to do non-interactive installs (ie. for something like a docker config) I've updated the README.md file with instructions for pre-answering the autostart question.

Beanow:
Hmm, that's unfortunate. We can keep that in mind for future improvements if people actually find it a problem. The interactive default is imo more important now.

Steve Myers <notifications@github.com> schreef op 13 september 2015 23:42:48 CEST:

mikehearn:
I was OK with the autostarting - it's pointless to install a Bitcoin node and then not actually use it. But I'm happy to delegate the Debian stuff to you, so will merge this now.

mikehearn:
Merged, thanks!

stevemyers:
Thanks Mike!  I'll keep an eye on feedback to see if we need to keep it this way.  

One nice thing about debconf is how ever you answer it won't ask you again (unless you removed the pkg first with a ```apt-get purge```).  So updates will always be non-interactive.

Issue #63, opened 2015-09-10
Rate limit the popup warnings?


Beanow:
This happened when I tried to send a transaction with one of the publicized private keys.
![Warning spam](http://ss.oscp.info/2015-09-10_05-09-55.png)

dgenr8:
Ah that's lovely. Can you add a few specifics?

Beanow:
Besides the UI thread being bricked completely for about 20 seconds at a time?
The wallet hadn't caught up with the mempool, so it thought inputs were still available, I made TXes from the Send tab, sent them and a ton of these came back over time.

I sent 2 TXes, resulting in all these alerts and also the failed TX from other people (using the same private key I assume) coming in.

dgenr8:
So that's what mom meant when she said there's no such thing as a free lunch ;)

zander:
I actually have been hit by a similar issue quite some time ago.  I'll take a look.

PR #73, opened 2015-09-14
Add a manager for the messages in the form of informationqueuedialog

zander:
All messages from any of our subcomponents would end up being
its own message box. This could cause some issues, as seen in the
latest "attack" where private keys were given away and people using
them would see hundreds of dialogs popping up about double spent
attacks.

This commit adds a manager in between that gathers up the messages
and sorts them and displays them in one dialog. In the special
case of the core spamming us with messages we just get the original
message updated with "repeated X times".

This addresses bugreport;
https://github.com/bitcoinxt/bitcoinxt/issues/63

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -82,6 +82,7 @@ QT_FORMS_UI = \
   qt/forms/editaddressdialog.ui \
   qt/forms/helpmessagedialog.ui \
   qt/forms/intro.ui \
+  qt/forms/informationqueuedialog.ui \
   qt/forms/openuridialog.ui \
   qt/forms/optionsdialog.ui \
   qt/forms/overviewpage.ui \
@@ -91,7 +92,8 @@ QT_FORMS_UI = \
   qt/forms/sendcoinsdialog.ui \
   qt/forms/sendcoinsentry.ui \
   qt/forms/signverifymessagedialog.ui \
-  qt/forms/transactiondescdialog.ui
+  qt/forms/transactiondescdialog.ui \
+  qt/forms/usermessagewidget.ui
 
 QT_MOC_CPP = \
   qt/moc_addressbookpage.cpp \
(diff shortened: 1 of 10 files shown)


dgenr8:
Impressive.  Got a screen shot?

mikehearn:
This is cool, but it'll take time for me to review it. Again, it'd be helpful if the part which isn't double-spends specific was upstreamed here: Jonas seems quite reasonable to me and it'd help cut down the rebasing cost if the Qt code doesn't diverge significantly. This is especially true because the Qt wallet isn't really competitive in many ways (e.g. no exchange rates support) and I don't think it has many users overall.

zander:
screenshot
screenshot1

zander:
The notifications are not really wallet specific..

I just noticed that core merged https://github.com/bitcoin/bitcoin/pull/6433 which makes Qt development utterly painful and is a classic example of throwing out the baby with the bathwater...

So if your intent is for XT to stay a patchset on top of core, then I'll close this request and we'll forget it.

jonasschnelli:
@zander why should https://github.com/bitcoin/bitcoin/pull/6433 make Qt development painful?

zander:
@jonasschnelli Do you think that
   foreach
is equally readable when you have to write
  Q_FOREACH
instead?

Notice that the IDE color highlights and autocompletes the former, not the latter.

Same with all the other Qt added keywords.

Its a bug in the buildsystem if this is actually a problem in the first place.

zander:
I see that Jonas wrote on the linked MR that this was definitely an improvement; on that we'll never agree.

jonasschnelli:
`Q_FOREACH` is perfectly fine and increase the readability of the code. It also makes it obvious which stuff is `BOOST` or `C++11` or `QT`. And last but not least it removes potential namespace conflicts.

IDE code highlight has IMO nothing todo with the code itself. Syntax highlighting is highly configurable and more flexible in cases of string-prefix namespaces like `Q_`

zander:
> IDE code highlight has IMO nothing todo with the code itself. Syntax highlighting is highly configurable and more flexible in cases of string-prefix namespaces like Q_

I understand your opinion, but unfortunately this is not shared by my IDE :)

PR #74, opened 2015-09-15
add setting to limit upload.

Meuh6879:
Hello,
In europe, we are "normally" limited to 1Mb upload.
but in villages, upload can be restrict to 256kb.

Can you add a setting like in the P2P file sharing software to "fix" an upload speed ?

This setting can be determined with a 1minute probe to test the maximum upload capacity with ping limit at 70ms or 100ms for example (noobs friendly automated setting).

thanks :)


mikehearn:
Hiya,

Feature requests like this are (for now) better put on the mailing list. We know that upload limits would be useful, but the issue is finding somebody to do it. By the way, where in Europe are you? I live in Europe and have much better than a 1mbit uplink, so does everyone else I know who lives here.

I'll close this for now as it's been opened as a pull request, but with no patches. So it'd be confusing to keep this open.

Meuh6879:
No problem, it's only a request to keep the data flow low for little connexion. ;)

The final can be a mix in a server machine with, on the left, the upload/download tools (emule, torrent, ftp content) and on the right, bitcoin with mining process to keep eyes on the bitcoin network activity.

Upload request in mining activity (hobbyist branch) is not low ... especially in stress phase of the network (clean the all input already spend and dust transaction).

PR #75, opened 2015-09-18
Add bandwith limiting to options-dialog

dagurval:
This adds bandwidth limiting elements to the Options-dialog in QT. This is based on the leaky bucket bandwidth limiter created by @gandrewstone, his work is cherry picked into this PR.

I think changing amt2Recv from int to int64_t also fixed an issue where I got disconnected from all peers if I disabled bandwidth limiting.

image

I think we need more sane defaults for bandwidth limiting, probably them off by default would be for the best.

Changing limits with RPC does not sync with GUI though. Not sure how to fix that. Do we need bandwidth limiting through RPC?

### DIFF
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -49,6 +49,12 @@ class CLeakyBucket
         }
         
 public:
+        // Initializes a disabled bucket.
+        CLeakyBucket() {
+            lastFill = clock.now();
+            level = max = fill = LONG_MAX;
+        }
+
         CLeakyBucket(int64_t maxp, int64_t fillp,int64_t startLevel=LONG_MAX):max(maxp),fill(fillp) 
         {
             lastFill = clock.now();            
@@ -65,10 +71,18 @@ public:
             lastFill = clock.now();  // need to reset the lastFill time in case we are turning on this leaky bucket.
         }
 
+        bool is_enabled() const {
+            return fill != LONG_MAX;
+        }
+
+        void disable() {
+            set(LONG_MAX, LONG_MAX);
+        }
+
         // Return the # tokens available if that amount is larger than the cutoff, otherwise return 0
         int64_t available(int64_t cutoff=0)
         {
-            if (fill == LONG_MAX) return LONG_MAX;  // shaping is off
+            if (!is_enabled()) return LONG_MAX;  // shaping is off
             fillIt();
             return (level > cutoff) ? level: 0;            
         }
@@ -76,7 +90,7 @@ public:
         // Try to use amt tokens.  Returns TRUE if the tokens were consumed, false otherwise
         bool try_leak(int64_t amt)
         {
-            if (fill == LONG_MAX) return true;  // leaky bucket is turned off.
+            if (!is_enabled()) return true;  // leaky bucket is turned off.
             assert(amt >=0);
             fillIt();
             if (level >= amt)
@@ -90,16 +104,12 @@ public:
         // This function reduces the level in the bucket by amt, even if that makes the level negative, and returns true if the level is >= 0.  This function is useful in a situation like data receipt (with soft limits) where you are not certain how many bytes will be received until after you have received them.
         bool leak(int64_t amt)
         {
-            if (fill == LONG_MAX) return true; // leaky bucket is turned off.
+            if (!is_enabled()) return true; // leaky bucket is turned off.
             fillIt();
             level-=amt;
             return (level>=0);
             
         }
-        
-        
-
-
 };
 
 #endif
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -1,0 +1,105 @@
+// Copyright (c) 2009-2013 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_LEAKYBUCKET_H
+#define BITCOIN_LEAKYBUCKET_H
+
+#if defined(HAVE_CONFIG_H)
+#include "config/bitcoin-config.h"
+#endif
+
+/** Default value for the maximum amount of data that can be received in a burst */
+static const int64_t DEFAULT_MAX_RECV_BURST = 100*1024; //*1024;
+/** Default value for the maximum amount of data that can be sent in a burst */
+static const int64_t DEFAULT_MAX_SEND_BURST = 100*1024; // *1024;
+/** Default value for the average amount of data received per second */
+static const int64_t DEFAULT_AVE_RECV = 30*1024; //*1024;
+/** Default value for the average amount of data sent per second */
+static const int64_t DEFAULT_AVE_SEND = 10*1024; // *1024;
+/** If we have to break the transmission up into chunks, this is the minimum send chunk size */
+static const int64_t SEND_SHAPER_MIN_FRAG = 256;
+/** If we have to break the transmission up into chunks, this is the minimum receive chunk size */
+static const int64_t RECV_SHAPER_MIN_FRAG = 256;
+
+
+class CLeakyBucket
+{
+    protected:
+        typedef boost::chrono::steady_clock CClock;
+    
+        int64_t level;  // Current level of the bucket
+        int64_t max;  // Maximum quantity allowed 
+        int64_t fill;  // Average rate per second
+        static CClock clock;
+        boost::chrono::time_point<CClock> lastFill;
+
+        // This function is called internally to fill the leaky bucket based on the time difference between now and the last time the function was called.
+        void fillIt()
+        {            
+          boost::chrono::time_point<CClock>  now = clock.now();
+          CClock::duration elapsed( now - lastFill);
+          int64_t msElapsed = boost::chrono::duration_cast<boost::chrono::milliseconds>(elapsed).count();
+          if (msElapsed > 100) // note in practice msElapsed can be < 0, something to do with hyperthreading so reduce don't eliminate this conditional
+          {
+              lastFill = now;
+              level += (fill*msElapsed)/1000;
+              if (level > max) level = max;
+          }        
+        }
+        
+public:
+        CLeakyBucket(int64_t maxp, int64_t fillp,int64_t startLevel=LONG_MAX):max(maxp),fill(fillp) 
+        {
+            lastFill = clock.now();            
+            // set the initial level to either what is specified by the user or to the maximum  
+            level = (startLevel < max) ? startLevel : max;
+        }
+
+        // Change the settings of the leaky bucket
+        void set(uint64_t maxp, uint64_t fillp)
+        {
+            max = maxp;
+            fill = fillp;
+            if (level > max) level=max;  // if pinching, slow traffic quickly.
+            lastFill = clock.now();  // need to reset the lastFill time in case we are turning on this leaky bucket.
+        }
+
+        // Return the # tokens available if that amount is larger than the cutoff, otherwise return 0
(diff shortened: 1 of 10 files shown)


rromanchuk:
Personally, i don't understand all these bandwidth tickets. I guess don't really understand who is running a full node and needs these controls.  If you're a user who has network limitations and do not want to run a light client, why... not just close QT?  If you're syncing, I can't see why you want this process to go slower. 

I also think these PRs make a lot more sense for core to be watered back at a later time.

jlopp:
If you want to understand the user's desire for bandwidth throttling, I suggest reading over the 4 year old issue in Core: https://github.com/bitcoin/bitcoin/issues/273

ptschip:
I don't really like the whole idea of bandwidth limiting, I'm not really sure how much it really helps the network other than to save the user from being able to browse the internet.  I agree that it will get more nodes up but I'm rather skeptical about what value those "bandwidth limited" nodes will bring  to the network as a whole.  I'm concerned that those nodes will end up slowing down propagation of blocks.  But then maybe not...I don't know.

That said, great work, and quickly too.  But I agree that the limiting should definitely be turned OFF by default.

mikehearn:
Nodes are useful for serving SPV clients even if they're at home.

ptschip:
Understood, but I would prefer if there were a way to not limit NEW
block propagation so that the miners can get their newly mined blocks to
propagate at the fastest speed through the network, and instead just do
the bandwidth limiting on transactions and older block requests from
users that are syncing new nodes.

JerryDurand:
I like that idea.  Always deal with the newest block first.  If your
network connection is really slow, of course you'll never see the older
blocks but that's not a fault of the software.

On 09/18/2015 03:55 PM, Peter Tschipper wrote:
> Understood, but I would prefer if there were a way to not limit NEW
> block propagation so that the miners can get their newly mined blocks to
> propagate at the fastest speed through the network, and instead just do
> the bandwidth limiting on transactions and older block requests from
> users that are syncing new nodes.
>

-- 
Jerry Durand, Durand Interstellar, Inc.  www.interstellar.com
tel: +1 408 356-3886, USA toll free: 1 866 356-3886
Skype:  jerrydurand

gandrewstone:
My motivation for making this patch is so I can just start-and-forget my bitcoin node.  Running raw XT I am seeing it interfere with skype and netflix (and I have fiber to the home).   With reasonable settings here netflix/skype video works fine and my node is still connected and communicating with lots of nodes.  Instant propagation verses nearly instant is really only critical for miners.

jonasschnelli:
Please don't add "dumb" bandwidth throttling to the bitcoin p2p network. A better approach would be prioritizing certain p2p messages or disable certain functions. Also a delayed inv'ing could make sense.
The `-maxuploadtarget` PR (https://github.com/bitcoin/bitcoin/pull/6622) does very effective limit the bandwidth without slowing down the p2p network. I did analyze certain nodes and did came to the point, that serving historical blocks for IBD makes up around 80% of the node traffic.

mikehearn:
@jonasschnelli If your patch makes sense, your explanation doesn't show how. That code 

a) Fails to work reliably, as noted in the flag description

b) Breaks mobile clients although they aren't the main user of bandwidth on the network

And as Pieter notes on the discussion, Bitcoin Core/XT should be able to easily handle slow peers that are throttled, as it downloads in parallel and stops using very slow peers.

jonasschnelli:
> a) Fails to work reliably, as noted in the flag description

Can you explain that?

> b) Breaks mobile clients although they aren't the main user of bandwidth on the network

No. It won't break mobile clients (unless they are only connected to one node).
Take an example: what if your mobile client is connected to a 10kb/s throttled node?
Node operators need to limit bandwidth in future. Especially if we increase the block size once. The question is, how would this affect the end user experience. Being connected to throttled nodes are always bad.

If your SPV client starts loading data from a node that likes to keep bandwidth low or reached a certain limit, you should go to the next node instead of loading data that is uncontrolled throttled.

> And as Pieter notes on the discussion, Bitcoin Core/XT should be able to easily handle slow peers that are throttled, as it downloads in parallel and stops using very slow peers.

For SPV clients this is extremely hard to detect. How would you detect if a node is throttled or your edge/3G/4G net is slow? Or other App consuming enough bandwidth?

My personal conclusion: throttling won't be a benefit in the bitcoin p2p world.

mikehearn:
> > a) Fails to work reliably, as noted in the flag description
>
> Can you explain that?

Well, it says in the patch description: "Currently there is no guarantee that the limit target will be fulfilled."

I think your patch is doing something different to this one. It appears to be a Tor style capped-quota limiter, for people who can use a max of N gigabytes of transfer per month. This patch doesn't let you do monthly/daily quotas but is intended to fix the issue of saturating limited bandwidth (megabits/sec).

For meeting the requirement of letting people watch streaming video at home, what's needed is a rate limiter like this one.

> Take an example: what if your mobile client is connected to a 10kb/s throttled node?

BitcoinJ switches to a faster one, if it needs to. If it gets the same speed after 3 switches it assumes the network itself is slow. It's fiddly but not impossible to do this sort of thing. But it shouldn't happen all that often.

Serving SPV wallets needs very little bandwidth overall, and is very important to end users. So restricting it whilst continuing to relay blocks and transactions makes little sense to me: the user will still go heavily over-quota, and there's less capacity to serve end users. Also just relaying transactions does not, by itself, do much for the network ... it just adds latency. Value to the P2P network comes from nodes that are used by merchants, that mine, that serve end users or which serve historical blocks to other nodes.

The breakage risk comes from someone deciding to helpfully set some default, then it stops being reconfigured upwards (see: soft/hard block size limits). And then as the network organically grows, one day that default gets hit and everyone is disconnected simultaneously.

MarcoFalke:
If you enable one of the "Limit *load" check boxes, shouldn't that disable incoming connections? Or at least scale the number of incoming/outgoing connections according to the selected bandwidth limit?

dgenr8:
@jonasschnelli If the network can't handle a throttled node, neither can it handle one that actually has less bandwidth.  I've never seen a plea for slow nodes not to connect.  Did I miss it?

jonasschnelli:
@dgenr8: not saying that the network can't handle throttled nodes. But throttled nodes are bad for the overall network quality.

There are better approaches to reduce overall bandwidth then just throttle the total net in/out of a Bitcoin node. Just copy/adapt the throttle mechanism of other p2p networks like BitTorrent is not well thought.

I think a node should reduce services or disconnect nodes if bandwidth gets critical.

In terms of other services getting disturbed by the Bitcoin node (if you watch a stream on Netflix as example), you should look at you operating systems or routers QoS features. Reducing bandwidth regardless of what else is going on on your system is a poor way of optimizing traffic.

I strongly recommend to avoid dumb network limiting. This can already be done with other tools/network shapers.

Bitcoin deserves a better way of handling limited bandwidth resources.

jlopp:
@jonasschnelli Yes, network limiting can be done better with more specialized tools, but the average Bitcoin user does not want to go to those lengths. If one of our goals is for there to be more nodes then we should make it as easy as possible to administer a node, which includes making it not disrupt the average user's residential Internet connection. 

I'm strongly for adding limiting features into the client itself; I agree that it can be improved more than this proposal. From my experience, the greatest bandwidth spikes my nodes have occur when nodes that are far behind on the blockchain request a large number of historical blocks. I'd suggest that serving large batches of historical blocks should be the first service that gets disabled in the event that a node decides it is using too much bandwidth.

dgenr8:
@jonasschnelli "Dumb" throttling has the advantage of being indistinguishable from an actual resource constraint.

"Smart" throttling, as you're defining it, involves reconsidering P2P design and tweaking incentives and balances.  Maybe those balances can be improved, but there is risk, and the improvements would need to be positive, regardless of whether they have a throttling effect.

There is a parallel to Gavin's mempool limiter.  Randomly kicking out txes has an effect that is essentially indistinguishable from the network growing unreliable.  Hence it works and got done quickly.  It doesn't stop the genius ideas from flowing in the future.  

There is another obvious parallel to increasing the block size, but I'll stop now.

gandrewstone:
@jonasschnelli: "dumb" throttling to X is equivalent to running a node
connected with bandwidth X, which is configurable.  Nodes with bandwidth X
can't be bad for the network, because that is all nodes.  "Throttling"
isn't bad,  but maybe throttling below a certain threshold is.  If you
believe that, you need to be numerically precise regarding what
send/receive bandwidths you think are "bad" for the network and back that
up with quantitative data.  I'd imagine that a lot of traffic shaping users
will have a LOT of bandwidth, and just want to bring things down to a
reasonable value.

But regardless, that is a different patch proposal: "Nodes should drop
connections with less bandwidth than N (irrespective of whether they are at
N because of throttling or slow internet)"

Secondly, I agree that a more intelligent throttling mechanism would be
better.  But my work is actually the basis for something like that.  Right
now the leaky bucket mechanism has a single "empty" level.  Adding
intelligent throttling would simply create additional levels in the
bucket.  For example you could stop sending certain classes of requests (or
reply to certain requests with an error) when the bucket is 1/2 full,
others at 1/8th etc.  But "smart" algorithms can often be tricked: its
still great to have a final cutoff, implemented simply and clearly, so that
the user's browsing or netflix experience is not diminished, causing him to
turn off the node.

JerryDurand:
Another point for bandwidth limiting is processing real time.  I've seen
on slower systems when catching up bitcoin can hog most of the
processing power causing the rest of the system to crawl.  Limiting the
bandwidth would reduce the workload and allow slow systems with a fast
connection to do other things.

-- 
Jerry Durand, Durand Interstellar, Inc.  www.interstellar.com
tel: +1 408 356-3886, USA toll free: 1 866 356-3886
Skype:  jerrydurand

mikehearn:
@JerryDurand probably not I'm afraid, given that a tiny download (e.g. 1mb) can generate a TON of disk and signature checking activity. If you want less load on the system during initial block download I think the right place to look is thread priorities and possibly adding optional sleep calls to various places in the code.

christophebiocca:
Limiting the total number of verification threads should work pretty well for limiting non-network resource usage, as long as the host system is multi-core itself.

    -par=<n>
    Set the number of script verification threads
    (-32 to 16, 0 = auto, <0 = leave that many cores free, default: 0)

dagurval:
For those concerned about throttling incoming connections - would dynamically reducing max connections based on limits be a good idea (as suggested by some)? Also very low limits could stop serving historical blocks.

Max connection could be reduced with block max block size. I haven't given thought into what a good rule would be, but as an example, a rule can be: Reduce max connections so that you could upload a full block to 1/8th of your clients at the same time within a minute.

One approach would be to have the client run recommended settings that does limit connections by default. To not hide this away from the user, we could add a "Advanced network options" dialog that shows current limits. Here the user could also override the settings as he/she likes.

I did a quick mock-up of what such a dialog could look like.

Thoughts?

image

mikehearn:
I think just having a basic bandwidth throttler is fine. Disabling serving of historical blocks is a solution to running out of transfer quota (X GB/month) but wouldn't fix streaming video/gaming applications.

gandrewstone:
Dagur,

Thanks for your code review.  You are entirely correct on all counts and
your description of how that system jams fake command line arguments into
the args object (the HORROR, the horror) was very helpful!  I think that I
addressed all of your comments in the latest checkin:

https://github.com/gandrewstone/bitcoinxt/commit/b7b139bb8c6accac1b1b616726e3fe1e7e9db036

Could you take a look?

Thanks!

mikehearn:
I believe this work has now been folded into the other pull reqs, so I will close this one. Please let me know if that's the wrong call.

PR #76, opened 2015-09-19

gandrewstone:
Includes QT GUI support.  The GUI support works fine but "stuff" is not perfect.  For example, the edit boxes are not aligned and too large relative to the sliders.  Also, I like how Dagur's example GUI says "kB/s".  Also it would be nice to have "validators".  I'm not sure if QT has these explicitly but basically they are a small routine that runs whenever the user makes a change in the edit box, and the box will (for example) be light pink if the value is impossible (text in a numeric field, or average > burst for example).

So in sum, I think that you might want to take this pull request since it is fully functional and then have someone better at QT than I am firm up my work there (or if there are no takers, I'll figure it out eventually).

### DIFF
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2009-2013 The Bitcoin Core developers
+// Copyright (c) 2015 The Bitcoin XT developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -55,6 +55,22 @@ public:
             // set the initial level to either what is specified by the user or to the maximum  
             level = (startLevel < max) ? startLevel : max;
         }
+
+        // Stop the operation of this leaky bucket (always return available tokens)
+        // use "set" to restart
+        void disable(void)
+	{
+  	  fill = LONG_MAX;
+          max = LONG_MAX;
+	}
+
+        // Access the values in this bucket
+        void get(uint64_t* maxp, uint64_t* fillp, uint64_t* levelp=NULL)
+	{
+	  if (maxp) *maxp = max;
+          if (fillp) *fillp = fill;
+          if (levelp) *levelp = level;
+	}
 
         // Change the settings of the leaky bucket
         void set(uint64_t maxp, uint64_t fillp)
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -1,0 +1,105 @@
+// Copyright (c) 2009-2013 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_LEAKYBUCKET_H
+#define BITCOIN_LEAKYBUCKET_H
+
+#if defined(HAVE_CONFIG_H)
+#include "config/bitcoin-config.h"
+#endif
+
+/** Default value for the maximum amount of data that can be received in a burst */
+static const int64_t DEFAULT_MAX_RECV_BURST = 100*1024; //*1024;
+/** Default value for the maximum amount of data that can be sent in a burst */
+static const int64_t DEFAULT_MAX_SEND_BURST = 100*1024; // *1024;
+/** Default value for the average amount of data received per second */
+static const int64_t DEFAULT_AVE_RECV = 30*1024; //*1024;
+/** Default value for the average amount of data sent per second */
+static const int64_t DEFAULT_AVE_SEND = 10*1024; // *1024;
+/** If we have to break the transmission up into chunks, this is the minimum send chunk size */
+static const int64_t SEND_SHAPER_MIN_FRAG = 256;
+/** If we have to break the transmission up into chunks, this is the minimum receive chunk size */
+static const int64_t RECV_SHAPER_MIN_FRAG = 256;
+
+
+class CLeakyBucket
+{
+    protected:
+        typedef boost::chrono::steady_clock CClock;
+    
+        int64_t level;  // Current level of the bucket
+        int64_t max;  // Maximum quantity allowed 
+        int64_t fill;  // Average rate per second
+        static CClock clock;
+        boost::chrono::time_point<CClock> lastFill;
+
+        // This function is called internally to fill the leaky bucket based on the time difference between now and the last time the function was called.
+        void fillIt()
+        {            
+          boost::chrono::time_point<CClock>  now = clock.now();
+          CClock::duration elapsed( now - lastFill);
+          int64_t msElapsed = boost::chrono::duration_cast<boost::chrono::milliseconds>(elapsed).count();
+          if (msElapsed > 100) // note in practice msElapsed can be < 0, something to do with hyperthreading so reduce don't eliminate this conditional
+          {
+              lastFill = now;
+              level += (fill*msElapsed)/1000;
+              if (level > max) level = max;
+          }        
+        }
+        
+public:
+        CLeakyBucket(int64_t maxp, int64_t fillp,int64_t startLevel=LONG_MAX):max(maxp),fill(fillp) 
+        {
+            lastFill = clock.now();            
+            // set the initial level to either what is specified by the user or to the maximum  
+            level = (startLevel < max) ? startLevel : max;
+        }
+
+        // Change the settings of the leaky bucket
+        void set(uint64_t maxp, uint64_t fillp)
+        {
+            max = maxp;
+            fill = fillp;
+            if (level > max) level=max;  // if pinching, slow traffic quickly.
+            lastFill = clock.now();  // need to reset the lastFill time in case we are turning on this leaky bucket.
+        }
+
+        // Return the # tokens available if that amount is larger than the cutoff, otherwise return 0
+        int64_t available(int64_t cutoff=0)
+        {
+            if (fill == LONG_MAX) return LONG_MAX;  // shaping is off
+            fillIt();
+            return (level > cutoff) ? level: 0;            
+        }
+        
+        // Try to use amt tokens.  Returns TRUE if the tokens were consumed, false otherwise
+        bool try_leak(int64_t amt)
+        {
+            if (fill == LONG_MAX) return true;  // leaky bucket is turned off.
+            assert(amt >=0);
+            fillIt();
+            if (level >= amt)
+            {
+                level-=amt;
+                return true;                
+            }
+            return false;            
+        }
+        
+        // This function reduces the level in the bucket by amt, even if that makes the level negative, and returns true if the level is >= 0.  This function is useful in a situation like data receipt (with soft limits) where you are not certain how many bytes will be received until after you have received them.
+        bool leak(int64_t amt)
+        {
+            if (fill == LONG_MAX) return true; // leaky bucket is turned off.
+            fillIt();
+            level-=amt;
+            return (level>=0);
+            
+        }
+        
+        
+
+
+};
+
+#endif
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -17,6 +17,7 @@
 #include "ui_interface.h"
 #include "crypto/common.h"
 #include "ipgroups.h"
+#include "leakybucket.h"
 
 #ifdef WIN32
 #include <string.h>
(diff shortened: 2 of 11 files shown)


ptschip:
Hi Dagur and Andrew,

I'm warming to the idea of bandwidth limiting however not without also
limiting the number of connected peers.  I don't see how the Bitcoin
network benefits if there are 125 peer connections on a node that are
then rate limited them after the fact?  Or an extreme example, if
someone made 5000 connections and then rate limited those connections to
a trickle, what would be the point? 

A real life example, I live here in a somewhat remote region.  I have
limited bandwidth and run with just 30 connections.  Everything runs
fine, I can browse the internet, I don't even know I have a node
running. Blocks and data are getting transmitted without delay to the
peers that are connected and connection slots on other peers are left
free and available to any node that that can service them better than
mine.  Now if someone were to start a new node and start downloading the
whole blockchain, or there was a spam attack, then yes, at that point I
could use some bandwidth limiting, and in fact I have had to knock off a
few nodes on occasion for that reason.  So at that point bandwidth
limiting would be of benefit, but not without being able to reduce the
number of connections also .   For that reason, I think it would be good
to include the ability to change the max connection setting through the
GUI as well as set the bandwidth limitations.  There could possibly be
guidelines available for new users, such as, if you have only 1Mbit/s
network then set to "X" connections with "Y" bandwidth limiting, etc...

ptschip:
Hi Peter,

I noticed an issue with my algorithm which is that the code iterates
through the connections in the same order each time.  While delays
throughout the iteration might cause the leaky bucket to fill mid-loop,
there is a delay at the bottom of the loop which causes the bucket to fill
at the end of the loop most of the time.

Originally I was going to fix this.  However, if you consider the effect
you will see that it basically prioritizes a consistent set of
connections.  So in a low-bandwidth situation, some connections will get no
(or very little) data.  The other side will see this as a non-responsive
connection and drop it.

So connection limiting is an emergent property of the system.

I verified this in an overnight run which had about 30 connections the next
morning.  Without limiting, I've seen over 100.

Of course, it would be better to run tests, etc.  However, the network as a
whole and a client in particular should not make assumptions about the
behavior of other clients.  Therefore, if we do find an issue the fix will
be on the other side, dropping low response as well as no-response
connections for example, not with the traffic shaper code.

tl;dr:  I think that we should take this bandwidth shaping code, start
using it, and make adjustments if needed to the other side to handle both
bandwidth shaping issues AND pathological behavior possibly caused by
network attacks.

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+
+#ifndef BITCOIN_LEAKYBUCKET_H
+#define BITCOIN_LEAKYBUCKET_H
+
+#if defined(HAVE_CONFIG_H)
+#include "config/bitcoin-config.h"
+#endif
+
+/** Default value for the maximum amount of data that can be received in a burst */
+static const int64_t DEFAULT_MAX_RECV_BURST = 100*1024; //*1024;
```
mikehearn: the //*1024 is confusing here, as the comments don't specify what the metric is. I'm assuming all these values are in bytes? Or does the comment imply they're measured in kilobytes?

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+static const int64_t DEFAULT_AVE_RECV = 30*1024; //*1024;
+/** Default value for the average amount of data sent per second */
+static const int64_t DEFAULT_AVE_SEND = 10*1024; // *1024;
+/** If we have to break the transmission up into chunks, this is the minimum send chunk size */
+static const int64_t SEND_SHAPER_MIN_FRAG = 256;
+/** If we have to break the transmission up into chunks, this is the minimum receive chunk size */
+static const int64_t RECV_SHAPER_MIN_FRAG = 256;
+
+
+class CLeakyBucket
```
mikehearn: As you have so diligently commented all the variables and methods, you may wish to add a class comment here stating exactly what the class does, perhaps link to the wiki article for the algorithm etc.

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+        int64_t level;  // Current level of the bucket
+        int64_t max;  // Maximum quantity allowed 
+        int64_t fill;  // Average rate per second
+        static CClock clock;
+        boost::chrono::time_point<CClock> lastFill;
+
+        // This function is called internally to fill the leaky bucket based on the time difference between now and the last time the function was called.
+        void fillIt()
+        {            
+          boost::chrono::time_point<CClock>  now = clock.now();
```
mikehearn: Indentation is quirky in this file. What editor are you using? You may wish to configure it to use 4 spaces consistently everywhere.

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+          if (msElapsed > 100) // note in practice msElapsed can be < 0, something to do with hyperthreading so reduce don't eliminate this conditional
+          {
+              lastFill = now;
+              level += (fill*msElapsed)/1000;
+              if (level > max) level = max;
+          }        
+        }
+        
+public:
+        CLeakyBucket(int64_t maxp, int64_t fillp,int64_t startLevel=LONG_MAX):max(maxp),fill(fillp)
```
mikehearn: Watch out for whitespace - again, please try to ensure it's consistent.

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+        // Stop the operation of this leaky bucket (always return available tokens)
+        // use "set" to restart
+        void disable(void)
+	{
+  	  fill = LONG_MAX;
+          max = LONG_MAX;
+	}
+
+        // Access the values in this bucket
+        void get(uint64_t* maxp, uint64_t* fillp, uint64_t* levelp=NULL)
```
mikehearn: What's wrong with just regular accessor functions, one for each? This pointer based API seems unnecessarily complex.

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+        void set(uint64_t maxp, uint64_t fillp)
+        {
+            max = maxp;
+            fill = fillp;
+            if (level > max) level=max;  // if pinching, slow traffic quickly.
+            lastFill = clock.now();  // need to reset the lastFill time in case we are turning on this leaky bucket.
+        }
+
+        // Return the # tokens available if that amount is larger than the cutoff, otherwise return 0
+        int64_t available(int64_t cutoff=0)
```
mikehearn: So if you call this method with the default argument, it'll always return zero? Why have a default argument here at all?

src/leakybucket.h
```diff
@@ -0,0 +1,121 @@
...
+            fillIt();
+            if (level >= amt)
+            {
+                level-=amt;
+                return true;                
+            }
+            return false;            
+        }
+        
+        // This function reduces the level in the bucket by amt, even if that makes the level negative, and returns true if the level is >= 0.  This function is useful in a situation like data receipt (with soft limits) where you are not certain how many bytes will be received until after you have received them.
```
mikehearn: Please wrap comments at something reasonable (e.g. 80 or 100 columns).

src/leakybucket.h
mikehearn: Sorry, ignore this comment.

src/net.h
```diff
@@ -49,6 +50,9 @@ static const int TIMEOUT_INTERVAL = 20 * 60;
 static const unsigned int MAX_INV_SZ = 50000;
 /** The maximum number of new addresses to accumulate before announcing. */
 static const unsigned int MAX_ADDR_TO_SEND = 1000;
+/** The maximum # of bytes to receive at once */
```
mikehearn: Perhaps this should remark that we allocate a buffer of this size on the stack, so making it too high can result in stack overflows? I do not know what the default stack size is on each platform but I'm guessing we wouldn't want this to be much larger than, say, 500kb.

src/net.cpp
```diff
@@ -950,14 +963,18 @@ void ThreadSocketHandler()
...
+                int64_t amt2Recv = receiveShaper.available(RECV_SHAPER_MIN_FRAG);
+                if (lockRecv&&(amt2Recv>0))
                 {
                     {
-                        // typical socket buffer is 8K-64K
-                        char pchBuf[0x10000];
+                        progress++;
+                        // max of min makes sure amt is in a range reasonable for buffer allocation
+                        int64_t amt = max((int64_t)1,min(amt2Recv,MAX_RECV_CHUNK));
+                        char pchBuf[amt];
```
mikehearn: I worry a bit about this being on the stack. It would be easy for it to accidentally end up too large, or having weird platform dependencies on operating systems that default to tiny stacks.

It might be better to move this onto the heap and access it via pnode instead.

src/net.cpp
```diff
@@ -1031,6 +1051,9 @@ void ThreadSocketHandler()
             BOOST_FOREACH(CNode* pnode, vNodesCopy)
                 pnode->Release();
         }
+
+        if (progress == 0)  // Nothing happened even though select did not block.  So slow us down.
+            MilliSleep(50);
```
mikehearn: 50 seems a bit arbitrary. This looks like it could busy wait a lot if the bandwidth throttler is set really low. Does it matter in practice? Maybe it should try and guess a good sleep value based on the throttle settings?

src/qt/optionsdialog.cpp
```diff
@@ -52,7 +59,7 @@ OptionsDialog::OptionsDialog(QWidget *parent, bool enableWallet) :
 
     ui->proxyIp->setEnabled(false);
     ui->proxyPort->setEnabled(false);
-    ui->proxyPort->setValidator(new QIntValidator(1, 65535, this));
+    ui->proxyPort->setValidator(&portValidator); //new QIntValidator(1, 65535, this));
```
mikehearn: What's this about? Looks like chaff left over from experimentation?

src/qt/optionsdialog.cpp
```diff
@@ -112,6 +119,52 @@ OptionsDialog::OptionsDialog(QWidget *parent, bool enableWallet) :
...
+    connect(ui->recvBurstSlider, SIGNAL(valueChanged(int)), this, SLOT(shapingSliderChanged()));
+    connect(ui->recvAveSlider, SIGNAL(valueChanged(int)), this, SLOT(shapingSliderChanged()));
+
+    connect(ui->recvAveEdit, SIGNAL(editingFinished()), this, SLOT(shapingAveEditFinished()));
+    connect(ui->sendAveEdit, SIGNAL(editingFinished()), this, SLOT(shapingAveEditFinished()));
+    connect(ui->recvBurstEdit, SIGNAL(editingFinished()), this, SLOT(shapingMaxEditFinished()));
+    connect(ui->sendBurstEdit, SIGNAL(editingFinished()), this, SLOT(shapingMaxEditFinished()));
+   
+    if (enabled)
+      {
```
mikehearn: Again, please fix your editor so it doesn't indent every line a random amount. Thanks!

src/rpcnet.cpp
```diff
@@ -165,6 +167,114 @@ Value getpeerinfo(const Array& params, bool fHelp)
...
+            "2. \"average\"  (integer, required) Specify the average throughput in Kbytes/sec\n"
+            "\nExamples:\n"
+            + HelpExampleCli("settrafficshaping", "\"receive\" 10000 1024")
+            + HelpExampleCli("settrafficshaping", "\"receive\" disable")
+            + HelpExampleRpc("settrafficshaping", "\"receive\" 10000 1024")
+        );
+
+    if (disable)
+      {
+        if (bucket) bucket->disable();
```
mikehearn: You've already checked bucket != NULL above.

src/rpcnet.cpp
```diff
@@ -165,6 +167,114 @@ Value getpeerinfo(const Array& params, bool fHelp)
...
+      }
+    else if (params.size() != 3) badArg = true;
+
+    if (fHelp || badArg || bucket==NULL)        
+        throw runtime_error(
+            "settrafficshaping \"send|receive\" \"burstKB\" \"averageKB\""
+            "\nSets the network send or receive bandwidth and burst in kilobytes per second.\n"
+            "\nArguments:\n"
+            "1. \"send|receive\"     (string, required) Are you setting the transmit or receive bandwidth\n"
+            "2. \"burst\"  (integer, required) Specify the maximum burst size in Kbytes/sec (actual max will be 1 packet larger than this number)\n"
```
mikehearn: I wonder if kiloBYTES/sec is really the right metric to specify here, given that bandwidth is normally measured in megaBITS/sec. Otherwise I guess nearly all real users will have to do the multiplication themselves. Seems a bit user unfriendly...

mikehearn:
Thanks for the patch! The code overall is looking very reasonable. We just need fixes for the following:

* I've got confused about what GUI code is currently best/recommended. Could you and dagurval comment on each pull request with whatever the latest code is meant to be?
* Whitespace! Check the patch in the github "files changed" view to see the issue :)
* I'm fairly sure the user should be specifying these caps in megabits/sec. I really doubt anyone running Bitcoin Core is on a <1 megabit/sec connection these days.
* Does it make sense to set the upstream cap a bit lower than what the user specifies in the Qt GUI? Because I suspect some users will say "I have a 10mbit/sec connection, so I'll set Bitcoin to 5mbit/sec for both" and then be upset when their TCP ACKs don't get through reliably.

I think the main UI usability questions here are around the users mental model of their asymmetric internet connection rather than sliders vs edits, etc.

I will be away for the next few days, speaking at a security conference in Romania. It'd be great if people could test this code and give reports like "I ran this for two days, watched Netflix with no problems and served lots of nodes historical blocks in this time".

PR #78, opened 2015-09-24

gandrewstone:
I think that this code is correct including the GUI (thanks to Dagur for help and feedback)!  Please test it if you have the opportunity.

I am still trying to find out where in github I can see the whitespace issues, so this pull request my not be acceptable WRT formatting.

Does anyone have a code formatter that (preferably emacs .el file) that formats properly for bitcoin-xt?

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -109,6 +109,7 @@ BITCOIN_CORE_H = \
   miner.h \
   mruset.h \
   net.h \
+  leakybucket.h \
   netbase.h \
   noui.h \
   policy/fees.h \
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2009-2013 The Bitcoin Core developers
+// Copyright (c) 2015 The Bitcoin XT developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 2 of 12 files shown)


gandrewstone:
I don't understand this travis-ci build failure:

net.h:22:25: fatal error: leakybucket.h: No such file or directory

leakybucket.h is a new file I added.  However it is in the same directory as net.h and included with quotes so it should be found... is it possible that travic-ci is not getting all the changes in the patch?

zander:
Travis seems to fail on many things, I wish they turned it off as long as it provides false failures.

gavinandresen:
@gandrewstone : you need to add leakybucket.h to Makefile.am .  The Travis build process first creates a source tarball (which will be missing the .h unless it is in Makefile.am) and then builds from that source.

@zander : Continuous build's benefits definitely outweigh the negatives, in my opinion. Travis has caught some chain-splitting bugs, and has caught lots of unportable code.

zander:
I found various compile errors on my 64 bit machine, related to the usage of the int64_t and uint64_t

I have to ask, can't we just use an int?

The config UI was not really working as I expected, the sliders didn't move when I edited the boxes and when using the slider (for burst) my text edited values were being changed...

I question the usage of sliders because its practically impossible to have good values for everyone. Some people like 100KB, some people want 10MB max.

I fixed some issues in a commit I pushed to my private repo;
https://github.com/zander/Bitcoin/commit/88144e0da4ce34dc0bec264c58a27c45fa451425

Feel free to cherry-pick from there.

gandrewstone:
I was mostly working on 32 bit.  We probably could just use int64 but need
to check LONG_MAX to make sure that does not overflow (I forget if it is
max signed or unsigned).    I made the slider change the dialog but not
vice versa (that could easily be added).  I like the sliders so lets see
what others think.  To fix the scale problem they could be exponential.
Unfortunately wont be online this weekend.

PR #80, opened 2015-10-01

gandrewstone:
Made the recommended changes.  Thanks for the reviews!  Let's see if it passes the ARM builds

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -109,6 +109,7 @@ BITCOIN_CORE_H = \
   miner.h \
   mruset.h \
   net.h \
+  leakybucket.h \
   netbase.h \
   noui.h \
   policy/fees.h \
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2009-2013 The Bitcoin Core developers
+// Copyright (c) 2015 The Bitcoin XT developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -55,6 +55,22 @@ public:
             // set the initial level to either what is specified by the user or to the maximum  
             level = (startLevel < max) ? startLevel : max;
         }
+
+        // Stop the operation of this leaky bucket (always return available tokens)
+        // use "set" to restart
+        void disable(void)
+	{
+  	  fill = LONG_MAX;
+          max = LONG_MAX;
+	}
+
+        // Access the values in this bucket
+        void get(uint64_t* maxp, uint64_t* fillp, uint64_t* levelp=NULL)
+	{
+	  if (maxp) *maxp = max;
+          if (fillp) *fillp = fill;
+          if (levelp) *levelp = level;
+	}
 
         // Change the settings of the leaky bucket
         void set(uint64_t maxp, uint64_t fillp)
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -10,13 +10,13 @@
 #endif
 
 /** Default value for the maximum amount of data that can be received in a burst */
-static const int64_t DEFAULT_MAX_RECV_BURST = 100*1024; //*1024;
+static const int64_t DEFAULT_MAX_RECV_BURST = 100 * 1024; //*1024;
 /** Default value for the maximum amount of data that can be sent in a burst */
-static const int64_t DEFAULT_MAX_SEND_BURST = 100*1024; // *1024;
+static const int64_t DEFAULT_MAX_SEND_BURST = 100 * 1024; // *1024;
 /** Default value for the average amount of data received per second */
-static const int64_t DEFAULT_AVE_RECV = 30*1024; //*1024;
+static const int64_t DEFAULT_AVE_RECV = 30 * 1024; //*1024;
 /** Default value for the average amount of data sent per second */
-static const int64_t DEFAULT_AVE_SEND = 10*1024; // *1024;
+static const int64_t DEFAULT_AVE_SEND = 10 * 1024; // *1024;
 /** If we have to break the transmission up into chunks, this is the minimum send chunk size */
 static const int64_t SEND_SHAPER_MIN_FRAG = 256;
 /** If we have to break the transmission up into chunks, this is the minimum receive chunk size */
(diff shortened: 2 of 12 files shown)


mikehearn:
Thanks for the updates. The formatting/whitespace problems are still there. You don't seem able to see them, I'm unsure why not. Please look at your code in the "files changed" view on github. Almost every line seems to have different indentation, there's no consistency in whether you put spaces around = signs or not, some lines are incredibly long (need wrapping) etc.

I can try and fix all these things for you, but whatever editor you're using seems to not understand C++ very well. I can suggest CLion from JetBrains. It's quite good.

I tried the code out. I noticed that bandwidth limits were enabled by default in the Qt UI even though I never asked for them. That seems wrong: they should be off by default.

The GUI does not explain what the limits are. It says they're measured in kilobytes, but that's not a bandwidth amount. I assume it means "KB/sec"? If so then the UI really needs to state that explicitly. And it should probably be measured in megabits/sec not kilobytes/sec as bandwidth is more normally measured in mbit/sec. I mentioned this before.

The two-slider UI was a bit unintuitive for me. I expected the average slider to have the max of whatever the max slider was set to, instead of showing some sort of error message if average > max. Why not just bind the average slider max value to the setting of the max slider?

gandrewstone:
Thanks for the feedback.  I'll work on your suggestions soon  Yes, I can't
see the spacing issues that you are seeing.  I am looking here:
https://github.com/bitcoinxt/bitcoinxt/compare/master...gandrewstone:master.
Can you give me a URL?  I will try CLion... BTW, what is the indentation
and line length standard for bitcoinxt?

With respect to megabit/sec vs kilobytes/sec, that would mean that the
smallest unit (without going into decimals) is 125 kilobytes/sec.  I think
that that is quite high -- I think that one group of people who use this
patch will be those with low bandwidth since they need to carefully manage
every bit of it.  And decimals might complicate the GUI and GUI logic...
How about kilobits/sec?

mikehearn:
For instance there are two-space indents in the first level of the fillIt() function, but 4 inside the if statement. If you look at disable() then the brackets are actually indented less than the function prototype, and "fill = LONG_MAX" is indented by one space less than the "max = LONG_MAX" line.

As another example you have code like:

    fill = fillp;
    if (level > max) level=max;

In the first line there's a space around the equals (correct) and in the if branch, there isn't.

I don't think there's a strict style guide for either Core or XT, but code should be consistent with itself. If you look at the rest of the code it tends to use 4 spaces to indent everywhere, spaces around equals, etc.

I realise this may seem very nitpicky, and I apologise for that, but it's a lot easier to read code when it isn't constantly changing code styles.

WRT megabits vs kilobits, are there really many users who have less than 1mbit/sec these days? Do they stand much chance of successfully bootstrapping a new node? I'm not against allowing more resolution (whether by decimals or not) but I'm not sure we should be optimising the UI for very slow connections. @gavinandresen what do you think?

zander:
My torrent app uses KiB/s, which makes most sense to me.

Usage of the KiB instead of KB just means you multiply by 1000 instead of 1024.

My UI proposal; 
options

mikehearn:
That UI looks pretty good Tom. Expressing burst as a percentage of the max seems better than requiring the user to slide it that way themselves. Most users probably wouldn't even adjust the default if it was a percentage.

gavinandresen:
People with very-low-bandwidth connections should not be running full nodes, they should be encouraged to run SPV.

RE: this pull: I think automatically limiting the number of connections if you're on a low-ish bandwidth connection is a better approach than slowing down traffic to all your peers.  Maybe if you're running the GUI it should default to making just 4 outgoing connections and allowing 8 incoming?  (assumption being GUI users are running from home, bitcoind users likely running in a datacenter)

You're asking me my opinion on a Monday as I'm getting over a cold, so please excuse the extra grumpiness; I don't care about the GUI (I think the Bitcoin-Qt/Bitcoin-Xt GUI is a terrible mess that just gets worse as programmers who never talk to end-users or perform usability testing with real people add Yet Another Option to it).

RE: code formatting:  the src/.clang-format file has all the rules for code formatting (in machine-readable-by-the-clang-format-tool format).

jonasschnelli:
Agree with @gavinandresen!

mikehearn:
Ah, great, so the clang-format tool can just auto fix everything.

I agree that Bitcoin-Qt does not have the most elegant or competitive wallet UI. But this feature exists for people who want to run it at home, so for them, some sort of UI is required even if it's not beautiful.

Reducing the number of connections wouldn't solve the actual problem of blocking upstream TCP ACKs whereas an actual bandwidth limiter like this one probably will. I'd still like to get feedback from people of the form "I watched Netflix all night whilst pumping blocks and it worked great", of course. We can ask on the mailing list for such feedback once test builds are available.

gandrewstone:
Nitpicky about spacing fine because I don't want to force you guys to do
menial labor.  My auto-formatter in emacs is different than what I saw in
bitcoinxt (and very aggressive) so I could not run it which is why you see
inconsistencies.

But looks like we *can* run an automatic formatting tool so I will do so
tonight.

I happen to be partial to sliders... but I do like Tom's burst percentage
as well.  And I'm fine with KiB/sec.  So I am going to just make the
quicker fixes to my GUI proposal and then let you guys decide which one you
want to accept.  If you choose Tom's, maybe just merge my full patch
(because unfortunately the GUI is not separate -- and anyway this way there
is a record of it) and then overwrite with his.

gandrewstone:
also WRT syncing given a slow connection: my intuition is that it is not
bandwidth limited.  I think that it is CPU/disk (validation) limited, and
also limited by the large timeout allowed for peers that send a partial
block.  I haven't done explicit testing on this, I've just watched the
bandwidth meter when syncing was in progress and sometimes bandwidth is
very low and little progress is being made...

As I was saying, I was throttling at modem speeds and had no problem
keeping in sync and supporting 20+ outgoing connections.  Take the patch
and experiment with it yourself!

gandrewstone:
Ok, just checked in code that should address all the issues.  I ran clang-format, spot checked a few places and trusted that it did the rest :-).

@gavinandresen : limiting the number of connections on low bandwidth connections will not solve the bursts which are what interferes with video chat and VOIP.  At the same time it may make sense to also limit # cnxns based on these settings so cnxns don't get starved of data.  However, my experiments show that some connection limiting is emergent behavior because the way I wrote the code is that the earlier nodes in the internal connection list always get all the bandwidth they can use.  But we may not want to rely on this since it means that some connections probably timed out which may inconvenience the receiving node (I haven't looked at the code there).

zander:
Hi.

first, thanks for your awesome work. This is great stuff!

the BitcoinXT branch you have been building on is the same as the BitcoinCore release branch. The next release split from it almost a year ago.
So if we want to re-use this work on top of the next core release, we have many many commits that have changed the underlying code to apply this on top of.

This means that a new feature should make minimal changes because any change that it makes is another potential conflict.

To further that goal I made a 2 commit version of your work that may prove to be much easier to merge and move forward to not lose it when we move to the next core release.
I want to make clear I build on top of your work, you did the heavy lifting :)

See merge request https://github.com/bitcoinxt/bitcoinxt/pull/81

It additionally makes the shaper use ints, like I suggested before and uses the UI that Mike seemed to prefer.

mikehearn:
Thanks @gandrewstone and @zander - how about we move discussion to Tom's split of this pull req?

dgenr8:
Close?

dgenr8:
Closing, traffic shaping has been live since 0.11D.

PR #81, opened 2015-10-07
Trafficshaping

zander:
Cleanup and squash the traffic-shaping commits to prepare them for fast-forward merging.

For easier review, I removed any changes that were not relevant to the functionality.

There is one 'issue' I noticed in that if you set the speed to a rather low speed (like 80KB/s up) the client takes 2 minutes to download a block, but the network layer (of my OS) says its done much much faster.  So maybe there is some buffering being done in this code that negates the effect on quite slow throughput.

Other than that, it works well. I've been running it for a day now with no ill effects.
Will update when I notice anything more.

### DIFF
--- a/src/leakybucket.cpp
+++ b/src/leakybucket.cpp
@@ -1,0 +1,83 @@
+// Copyright (c) 2015 The Bitcoin XT developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "leakybucket.h"
+
+CLeakyBucket::CLeakyBucket(int maxp, int fillp, int startLevel)
+    : max(maxp),
+    fill(fillp)
+{
+    lastFill = clock.now();
+    // set the initial level to either what is specified by the user or to the maximum
+    level = (startLevel < max) ? startLevel : max;
+}
+
+void CLeakyBucket::disable()
+{
+    fill = 0;
+    max = 0;
+}
+
+void CLeakyBucket::fillIt()
+{
+    boost::chrono::time_point<CClock> now = clock.now();
+    CClock::duration elapsed(now - lastFill);
+    int msElapsed = boost::chrono::duration_cast<boost::chrono::milliseconds>(elapsed).count();
+    // note in practice msElapsed can be < 0, something to do with hyperthreading
+    // so don't eliminate this conditional
+    if (msElapsed > 100)
+    {
+        lastFill = now;
+        level += (fill * msElapsed) / 1000;
+        if (level > max)
+            level = max;
+    }
+}
+
+void CLeakyBucket::get(int* maxp, int* fillp, int* levelp)
+{
+    if (maxp) *maxp = max;
+    if (fillp) *fillp = fill;
+    if (levelp) *levelp = level;
+}
+
+void CLeakyBucket::set(int maxp, int fillp)
+{
+    max = maxp;
+    fill = fillp;
+    if (level > max)
+        level=max; // if pinching, slow traffic quickly.
+    lastFill = clock.now(); // need to reset the lastFill time in case we are turning on this leaky bucket.
+}
+
+int CLeakyBucket::available(int cutoff)
+{
+    if (fill == 0)
+        return INT_MAX; // shaping is off
+    fillIt();
+    return (level > cutoff) ? level : 0;
+}
+
+bool CLeakyBucket::try_consume(int amt)
+{
+    if (fill == 0)
+        return true; // leaky bucket is turned off.
+    assert(amt >= 0);
+    fillIt();
+    if (level >= amt)
+    {
+        level-=amt;
+        return true;
+    }
+    return false;
+}
+
+bool CLeakyBucket::consume(int amt)
+{
+    if (fill == 0)
+        return true; // leaky bucket is turned off.
+    fillIt();
+    level -= amt;
+    return (level >= 0);
+}
--- a/src/leakybucket.h
+++ b/src/leakybucket.h
@@ -1,0 +1,64 @@
+// Copyright (c) 2015 The Bitcoin XT developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_LEAKYBUCKET_H
+#define BITCOIN_LEAKYBUCKET_H
+
+#include <boost/chrono/include.hpp>
+
+/** Default value for the maximum amount of data that can be received in a burst */
+static const int DEFAULT_MAX_RECV_BURST = 100 * 1024;
+/** Default value for the maximum amount of data that can be sent in a burst */
+static const int DEFAULT_MAX_SEND_BURST = 100 * 1024;
+/** Default value for the average amount of data received per second */
+static const int DEFAULT_AVE_RECV = 30 * 1024;
+/** Default value for the average amount of data sent per second */
+static const int DEFAULT_AVE_SEND = 10 * 1024;
+/** If we have to break the transmission up into chunks, this is the minimum send chunk size */
+static const int SEND_SHAPER_MIN_FRAG = 256;
+/** If we have to break the transmission up into chunks, this is the minimum receive chunk size */
+static const int RECV_SHAPER_MIN_FRAG = 256;
+
+class CLeakyBucket
+{
+public:
+    CLeakyBucket(int maxp, int fillp, int startLevel = INT_MAX);
+
+    // Stop the operation of this leaky bucket (always return available tokens)
+    // use "set" to restart
+    void disable();
+
+    // Access the values in this bucket
+    void get(int* maximum, int* fillAmount, int *current = NULL);
+
+    // Change the settings of the leaky bucket
+    void set(int maximum, int fillAmount);
+
+    // Return the # tokens available if that amount is larger than the cutoff, otherwise return 0
+    int available(int cutoff = 0);
+
+    // Try to use \a amount tokens.  Returns true if the tokens can be consumed, false otherwise
+    bool try_consume(int amount);
+
+    // This function reduces the level in the bucket by amt, even if that makes the
+    // level negative, and returns true if the level is >= 0. This function is useful
+    // in a situation like data receipt (with soft limits) where you are not certain
+    // how many bytes will be received until after you have received them.
+    bool consume(int amount);
+
+protected:
+    typedef boost::chrono::steady_clock CClock;
+
+    int level; // Current level of the bucket
+    int max; // Maximum quantity allowed
+    int fill; // Average rate per second
+    static CClock clock;
+    boost::chrono::time_point<CClock> lastFill;
+
+    // This function is called internally to fill the leaky bucket based on
+    // the time difference between now and the last time the function was called.
+    void fillIt();
+};
+
+#endif
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -108,6 +108,11 @@ boost::condition_variable messageHandlerCondition;
 static CNodeSignals g_signals;
 CNodeSignals& GetNodeSignals() { return g_signals; }
 
+// Variables for traffic shaping
+CLeakyBucket receiveShaper(0 ,0);
+CLeakyBucket sendShaper(0, 0);
+boost::chrono::steady_clock CLeakyBucket::clock;
+
 void AddOneShot(string strDest)
 {
     LOCK(cs_vOneShots);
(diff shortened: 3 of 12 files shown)


mikehearn:
Thanks Tom.

@gandrewstone WRT to starving connections depending on where they are in the connection list. I agree that peers will handle this (due to the pings they do) but it's perhaps not ideal. Is there a simple way to make it fairer?

src/leakybucket.cpp
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2015 The Bitcoin XT developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "leakybucket.h"
+
+void CLeakyBucket::fillIt()
+{
+    boost::chrono::time_point<CClock>  now = clock.now();
+    CClock::duration elapsed( now - lastFill);
```
mikehearn: whitespace on this line and the one above (inconsistent spacing)

mikehearn:
Actually, I just noticed that Andrew's last commit introduced quite a few fixes and changes. It seems now this change has been forked and someone will need to merge .... :(

zander:
I went though the new class and did the whitespace fixes & forcepushed.

I looked at the commit from yesterday from gandrewstone, none of the changes were relevant for these commits. You can consider them merged :)

zander:
Just wanted to add some more experiences; the current spamattack is actually useful for getting proper results.
Here is a screenshot where the limits are set to 250KB/s.  The image shows at the top the Bitcoin clients perceived notion of the world.  The bottom is what the OS thinks.  The x axis is not entirely equal, but you can easy make out the plateaus and max speeds.  Which means that the concept really works quite well.
Spikes are because I set burst to 300% for outgoing and 1000% for incoming.
performance

gandrewstone:
Its kind of rude to recommit my changes under your own name.  Its
unfortunate that the changes are based off the wrong branch but you had
lots of time to tell me that while you were nit-picking about spacing and
integer size issues.  If this is the way you recognize contributions, I am
done here.

Regards,
Andrew

gavinandresen:
@gandrewstone : the commit messages clearly recognize you, please don't be so sensitive (or, if you can't help being really sensitive about exactly how people are recognized for their contributions, then yes, this isn't the project for you; in my opinion, we don't need that kind of drama).

zander:
> Its kind of rude to recommit my changes under your own name. 

Did you notice my commit message:
```
Add a command 'settrafficshaping' and 'gettrafficshaping' to the
RPC interface.

Based on code from
    dagurval@gmail.com and
    G Andrew Stone <g.andrew.stone@gmail.com>
```

>  Its unfortunate that the changes are based off the wrong branch but you had lots of time to tell me that while you were nit-picking about spacing and integer size issues

I'm sorry if I didn't explain this; its not "the wrong branch".   The XT client doesn't really have the luxury of having its own branch. Its a stack of commits on top of Core. Or as others called it "core++"

Please be fair, I have mentioned the integer and spacing issues on various occasions, including my very first reply to your work (on the ML).

mikehearn:
Hey Andrew,

We absolutely will recognise your contributions! The commits can have the author reset, I fully understand why you're annoyed about that. I think it's not intentional but rather an artifact of how git works.

For what it's worth, in the XT release announcements and patches page, I always call out the primary contributor of a feature which in this case would be you. I do this for bitcoinj too, look:

   https://bitcoinj.github.io/release-notes

Credit where it's due is important.

ptschip:
On another note, I've don't know if anybody has compiled and run the
latest on Windows but it works fine!  First time I've tried it out and
am impressed that it works so well.

I'd still like to see a way to dynamically change the number of
connected nodes but I'm not complaining.  That can always be done
another time.

gandrewstone:
@gavin and @mike.  Why don't you guys IM each other to figure out how
important properly recording contributions is going to be for this project,
make the changes in github (if any), and then let me know?

Regards,
Andrew

mikehearn:
We both agree attribution is important, the question is just about the git author line (i.e. the little face in github). I am happy to edit things back so it's your avatar that shows up there, if you like. Then Thomas' work would be recognised in the commit message instead.

zander:
@gandrewstone 

I was trying to help you by doing the boring stuff and avoiding having lots of comments and revisions.

I assumed we are all here working on this project to have a better product. Not to have a commit somehow in your name in one specific way...
People reviewing spent hours on that and never get acknowledged in commit messages or authors, this is open source where we build something *together*.
Why the drama?

The commitmessage gives credit already, I suggest we review these commits and merge them so we can ship this soon.

gandrewstone:
Clearly I differ philosophically from some of you (but not Mike, it seems)
WRT the importance of correct attribution.  So maybe this isn't the project
for me or maybe I'll just suck it up :-).

@tomz There are probably lots of ways to preserve commit history but since
the pull was targeted to the wrong branch anyway I would have just grabbed
the pull and then merged it into the other branch (with your changes).
That may have been even faster then the manual technique you employed.

@gavin Maybe you should ask your wife (who is a professor I read somewhere)
her opinion on correct attribution.

Regardless, to be completely clear, my small submission remains licensed
under the COPYING file (essentially MIT licensed) so feel free to
incorporate it in whatever way you want.

Regards,
Andrew

mikehearn:
@zander I didn't understand your comment about the changes from yesterday not being relevant. For instance Andrew fixed the issue with the limiter being on by default. But this pull still has it that way. It seems the changes should be merged in, or his current patch re-split?

zander:
Hi Mike,

the comment I made yesterday was maybe a bit too fast. Certainly too terse.
I read through the code and concluded that the code changes could not be merged because it was fixing issues merge #80 had of various sizes of integers while this one uses only int.
The change to change the defaults of all the shaping to MAX is similarly not applicable because in this merge the "off" value is zero, not max.  Additionally, changing the const ints has the negative side effect of making the settings UI not have nice defaults any longer.
I think code-style fixes made were of code that wasn't originally touched by this new feature.


The intention of the patch, however, is something that should indeed be merged.  While the bitcoin-qt had the shaper off by default, bitcoind didn't.  I'll push a change for that shortly.

Since you didn't really start commenting yet, I'll force-push. The changes are in net.cpp where I stopped using the const ints for the defaults to getArg and just supply zero, which is equal to off.
This disconnect is due to a difference in the UI where a 3rd variable (bool) is used for turning the shaper on/off.

So, to be clear, I didn't just squash and split Andrews code.
I changed the shaper class to not be pure inline any longer.
I changed the shaper to use ints.
I changed the shaper to use zero as its default value.
I changed the UI to my own design. This dropped a lot of unneeded validation code.
I removed the initialization code in the UI which was useless.
I reverted any changes  that were not strictly applicable to the feature.
I formatted many changes to fit in our coding style.
I fixed some gcc-isms.
Fixed the copyright to be how FSF wants it.
changed the RPC to be more consistent with the other commands.
Used "KB/s" instead of "KBytes/s"
Actually used KB instead of KiB in the code (also for other code consistency).
Removed 'external' declaration from a net.h, which looked to intrusive to me.

The intent was to get higher quality code, and lower future collisions in merge/rebase.  I tried a cherry-pick on core-master and apart from some refactor in the RPC, there was nothing really problematic. So I have hope that this merge request can be forward ported when that day comes.

gandrewstone:
Just in case there is any question about how easy it is to provide proper attribution using git (and to help git beginners) I have reviewed and added a few small changes to the dynamic spam blocker, preserving ptschip's commit history:  https://github.com/gandrewstone/bitcoinxt/commits/spamblocker and you can see the pull request here https://github.com/ptschip/bitcoinxt/pull/1.  I did not have to submit my pull request to ptschip/bitcoinxt.  I could just as easily have submitted the pull request to bitcoinxt/bitcoinxt.  But I am submitting my changes to him so that he can review my review.  

These are the steps for those who are unfamiliar with the awesome P2P nature of git:

In my fork:
git remote add ptschip git@github.com:ptschip/bitcoinxt.git  (tell git about this other repo)
git fetch ptschip (grab his branches)
git checkout spamblocker  (switch to where the code is)
(now make changes and test)
git commit -a -m "my changes"
git push origin spamblocker (push this branch up to my github repo)

Then go to the github GUI and generate a pull request as normal.

mikehearn:
OK, thanks Thomas, and don't worry Andrew, I fully intend to give you the attribution in the git authors field (with of course some credit to Thomas too in the commit description for his contributions). I'll also name everyone who chipped in, in the release notes.

src/leakybucket.h
```diff
@@ -0,0 +1,64 @@
...
+    void get(int* maximum, int* fillAmount, int *current = NULL);
+
+    // Change the settings of the leaky bucket
+    void set(int maximum, int fillAmount);
+
+    // Return the # tokens available if that amount is larger than the cutoff, otherwise return 0
+    int available(int cutoff = 0);
+
+    // Try to use amt tokens.  Returns TRUE if the tokens were consumed, false otherwise
+    bool try_leak(int amount);
```
mikehearn: "use" would be a better name than "leak"

src/leakybucket.h
```diff
@@ -0,0 +1,64 @@
...
+
+    // This function reduces the level in the bucket by amt, even if that makes the
+    // level negative, and returns true if the level is >= 0. This function is useful
+    // in a situation like data receipt (with soft limits) where you are not certain
+    // how many bytes will be received until after you have received them.
+    bool leak(int amount);
```
mikehearn: Same here. Or maybe "consume". I understand that leak is appropriate in the context of the leaky bucket analogy, but it just looks confusing at the call site IMO

src/net.cpp
```diff
@@ -1031,6 +1060,9 @@ void ThreadSocketHandler()
             BOOST_FOREACH(CNode* pnode, vNodesCopy)
                 pnode->Release();
         }
+
+        if (progress == 0)  // Nothing happened even though select did not block.  So slow us down.
+            MilliSleep(50);
```
mikehearn: It'd be good to have a rationale in the comments for why 50 msec here.

src/net.h
```diff
@@ -49,6 +50,9 @@ static const int TIMEOUT_INTERVAL = 20 * 60;
 static const unsigned int MAX_INV_SZ = 50000;
 /** The maximum number of new addresses to accumulate before announcing. */
 static const unsigned int MAX_ADDR_TO_SEND = 1000;
+/** The maximum # of bytes to receive at once */
+static const int MAX_RECV_CHUNK = 256*1024;
```
mikehearn: Isn't this the same as RECV_SHAPER_MIN_FRAG?

mikehearn:
This is looking pretty good now - nearly there! One more round of minor tweaks and I'll spin some test builds and ask for wider feedback.

src/net.cpp
ptschip: I'm not sure I see the rational for the Millisleep(50) here.   Select
will block until it's timeout is reached.  If there's nothing to do it
will just go back to select.  Why the need for an additional sleep?

gandrewstone: Select won't block because there is data available.  But we are not "harvesting" it because there is not enough room in the leaky bucket.

50msec is unfortunately completely arbitrary.

ptschip: perhaps you could turn off the 50 millisecond sleep when rate limiting
is not turned on so that it doesn't impact the performance for those
wanting to rate limit...

src/net.h
gandrewstone: No they are different.

RECV_SHAPER_MIN_FRAG is basically saying "if the bucket is limiting us to less the X bytes, then don't bother to read anything at all (because its inefficent to read 1 byte at a time)" -- that is what is the minimum fragment we should read?  It is currently set to 256 bytes."

MAX_RECV_CHUNK is saying "Even if the leaky bucket is allowing us to receive 10MB, only grab X bytes in this loop iteration".  This was a hard-coded value (size of the buffer allocated on the stack) in the original code.  It is set to 256K bytes.

src/net.cpp
gandrewstone: If rate limiting is off, data will always be read so "progress" will be
nonzero.  So this sleep will "never" occur if rate limiting is off.

"Never" is in quotes because the only time it would sleep is if there is
some kind of bug where nothing is read.  Presumably this means that the
looping call to select() will never block because the code fails to consume
any data.  In this case bitcoind would hit 100% CPU (if this sleep code was
not there).  So this code "helps" the non-shaping case by keeping the
system usable even if there is a bug in the select handling code.

Of course, any code that reduces the effects of bugs may also have the side
effect of hiding them... so you can collectively decide whether you want a
compound if statement, also checking whether the leaky bucket is on, here.

zander:
Renamed leak to consume.

Updated the 50ms comment.

Fixed author field.

mikehearn:
I've merged this work into master, as nobody seemed to have complaints it's time for wider testing.

Big thanks to G Andrew Stone and Thomas Zander for their efforts.

PR #83, opened 2015-10-09
Bump UPnP version as the current one has a buffer overflow in it

mikehearn:


ptschip:
@mikehearn  is this the same version that Core is bumping up to?  (v1.9.20151008)

gavinandresen:
ACK; lightly tested on OSX (updated miniupnpc : brew unlink/edit/install, then compiled and ran).

PR #84, opened 2015-10-09
Dynamically Adjusting Spam Blocker

ptschip:
As the memory pool fills up to more than one
block height the limitfreerelay and minrelaytxfee
are dynamically adjusted to gradually choke off low fee spam.  

The minrelaytxfee is used as an upper bounds and is gradually increased from zero to the minrelaytxfee value as the choking is invoked.  limitfreerelay is also gradually choked off but will not go below 5.

If/when the spamming ends and the memory pool begins to fall in size the choke is also gradually removed.

Spam transaction are never accepted into the memory pool and therefore not
relayed stopping the spam from traversing the network, saving CPU , memory and bandwidth.

Currently it will reach maximum choke after a 3 block height in the memory pool but even then 
some free transactions will still get through.

Awaiting comments...

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1012,7 +1012,7 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
     if (pfMissingInputs)
         *pfMissingInputs = false;
 
-    int64_t maxBlockSize = Params().GetConsensus().MaxBlockSize(GetAdjustedTime(), sizeForkTime.load());
+    uint64_t maxBlockSize = Params().GetConsensus().MaxBlockSize(GetAdjustedTime(), sizeForkTime.load());
 
     if (!CheckTransaction(tx, state, maxBlockSize))
         return error("AcceptToMemoryPool: CheckTransaction failed");
@@ -1162,7 +1162,7 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         }
         else if(poolBytes < maxBlockSize * MAX_BLOCK_SIZE_MULTIPLYER){
             // Gradually choke off what is considered a free transaction
-            feeCutoff = (minRelayTxFee/1000) * (poolBytes-maxBlockSize)/(maxBlockSize * (MAX_BLOCK_SIZE_MULTIPLYER-1));
+            feeCutoff = (minRelayTxFee/1000) * (poolBytes - maxBlockSize) / (maxBlockSize * (MAX_BLOCK_SIZE_MULTIPLYER-1));
 
             // Gradually choke off the nFreeLimit as well but leave at least minFreeLimit
             // So that some free transactions can still get through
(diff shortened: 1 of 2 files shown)


rromanchuk:
I really like this concept. I *want* to keep substandard txfee transactions in my mempool. I want to give good actors a chance to participate in the network. It's up to the miners to decide at which fee the market clears. "spam" attacks from bad actors seem pretty easy to detect, through the abnormal increase in mempool size. This obviously still drops good actors, but a much better start than manually adjusting minrelay up and down.  Unfortunately, I have a feeling a majority of the nodes that raised their minrealy will probably now leave them there indefinitely which prevents good actors from participating.

zander:
You call it "spam" but all I see is "low fee" as what you are blocking.

A user sending a valid low fee transaction should not be having problems getting his transaction accepted just because the network is busy.

I feel this approach misses the advantage of distributed computing where thousands of nodes each have a part of the pool.  It assumes there is only one pool and we need to protect it.  But thats not how it works. The actual capacity of the bitcoin network for unconfirmed transactions is many many times one block.

The goal should be to avoid losing transactions, not to block ones that you programatically think is "spam".

zander:
The current solution, and the improvement Jonathan is working on, for mempool eviction solves this problem in a proper distributed computing (i.e. massive redundancy) way that scales much better and doesn't discriminate.

ptschip:
On 09/10/2015 2:15 PM, TomZ wrote:
>
> You call it "spam" but all I see is "low fee" as what you are blocking.
>
I think that all this code is doing is automating those two features
that are already in the code.  With the benefit that it "almost"
completely stops the movement of spam (or low fee) transactions through
the network.  I don't see how any spammer would bother making the
attempt...there is nowhere for their transactions to go.


> A user sending a valid low fee transaction should not be having
> problems getting his transaction accepted just because the network is
> busy.
>

As far as how many blocks height the mempool can be, that is easily
configurable.  I made it purposly aggressive to test it out, but we
could easily leave 10 or 20 blocks full.
>
> I feel this approach misses the advantage of distributed computing
> where thousands of nodes each have a part of the pool. It assumes
> there is only one pool and we need to protect it. But thats not how it
> works. The actual capacity of the bitcoin network for unconfirmed
> transactions is many many times one block.
>
I think if everynode used this approach we wouldn't have any spammers to
deal with and wouldn't have to worry about what gets in out of the mempool.
>
> The goal should be to avoid losing transactions, not to block ones
> that you programatically think is "spam".
>
Actually I think more free transactions are able to get through with
this approach.  No low fee or free transactions are stopped at all until
we get a full memory pool (or wherever we decide the choke should
be)...whereas now in the current code, we have a default values  of
minrelaytxfee = 0.00001501
 that can prevent free transactions from getting into the mempool before
it's even full.

ptschip:
I added slightly better logging for anyone that wants to run this and
see it in action.  I've been running it for 4 days now and haven't seen
the memory pool got higher than 5MB.  Mostly it's cylcing around 2.5 to
4MB during this low fee spam attack.

Also, my intention here is not to replace the work that's already been
done with pruning the mempool but rather to create an additional first
hurdle that the spammers have to get over.  I see both approaches as
complimentary and necessary safeguards.

You said that " I have a feeling a majority of the nodes that raised
their minrealy will probably now leave them there indefinitely which
prevents good actors from participating"...however the way this code is
written is that the minrelay always stays at zero "until" the memory
pool is full, no matter what value has been set in the config file.

mikehearn:
I think design decisions like this are best debated on the mailing list rather than github. Github is best used for discussing specifics of code changes rather than the overarching design, IMO.

I'll reply on the list, please follow me there.

ptschip:
I'll close this out for now...Doesn't seem like much interest and it looks like the Core team is getting close to merging their own version.

PR #89, opened 2015-10-28
evictRandomBytewise with worst-of-2 eviction

jtoomim:
This PR adds the ability to limit the mempool size to a number of bytes, using the -maxmempoolbytes command line option. It adds a method to pick two random transactions from the mempool, weighted by the transactions' sizes, and evict the transaction with the lowest fee/kB. Transactions that are prioritized will be skipped.

The -maxmempoolbytes option defaults to 50 blocks' worth of transactions. It uses the transaction's serialized size, not the in-memory size. The in-memory size is typically 2.5x higher. Thus, with 1 MB blocks, the default setting allows for approximately 2.5 * 50 * 1 MB = 125 MB of memory usage by the mempool.

Performance-wise, each eviction typically takes less than 200 us on a Core i7 4790k. Performance should be roughly O(1) vs mempool size.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -314,7 +314,8 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-maxconnections=<n>", strprintf(_("Maintain at most <n> connections to peers (default: %u)"), 125));
     strUsage += HelpMessageOpt("-maxreceivebuffer=<n>", strprintf(_("Maximum per-connection receive buffer, <n>*1000 bytes (default: %u)"), 5000));
     strUsage += HelpMessageOpt("-maxsendbuffer=<n>", strprintf(_("Maximum per-connection send buffer, <n>*1000 bytes (default: %u)"), 1000));
-    strUsage += HelpMessageOpt("-maxmempooltx=<n>", ("Maximum number of transactions in memory pool (default: enough to fill about 25 blocks)"));
+    strUsage += HelpMessageOpt("-maxmempooltx=<n>", ("Maximum number of transactions in memory pool (default: enough to fill about 25 blocks with 500 byte transactions)"));
+    strUsage += HelpMessageOpt("-maxmempoolbytes=<n>", ("Maximum size (in bytes) of transactions in memory pool (default: enough to fill 50 blocks)"));
     strUsage += HelpMessageOpt("-onion=<ip:port>", strprintf(_("Use separate SOCKS5 proxy to reach peers via Tor hidden services (default: %s)"), "-proxy"));
     strUsage += HelpMessageOpt("-onlynet=<net>", _("Only connect to nodes in network <net> (ipv4, ipv6 or onion)"));
     strUsage += HelpMessageOpt("-permitbaremultisig", strprintf(_("Relay non-P2SH multisig (default: %u)"), 1));
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -315,7 +315,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-maxreceivebuffer=<n>", strprintf(_("Maximum per-connection receive buffer, <n>*1000 bytes (default: %u)"), 5000));
     strUsage += HelpMessageOpt("-maxsendbuffer=<n>", strprintf(_("Maximum per-connection send buffer, <n>*1000 bytes (default: %u)"), 1000));
     strUsage += HelpMessageOpt("-maxmempooltx=<n>", ("Maximum number of transactions in memory pool (default: enough to fill about 25 blocks with 500 byte transactions)"));
-    strUsage += HelpMessageOpt("-maxmempoolbytes=<n>", ("Maximum size (in bytes) of transactions in memory pool (default: enough to fill 50 blocks)"));
+    strUsage += HelpMessageOpt("-maxmempoolbytes=<n>", ("Maximum in-memory size of transactions in memory pool (default: enough to fill 50 blocks)"));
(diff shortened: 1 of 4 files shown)


zander:
Code looks good! unACK.

gavinandresen:
+1 to removing the timing code-- if I recall correctly, some versions of boost on some platforms have issues with nanosecond-resolution clocks.

I'd also vote to remove the -debug=mempool detail on which candidates are compared; I'm sure it was useful during development, but I doubt it will ever be useful again now the code is written and debugged.

Other than that untested ACK.

dgenr8:
I'm pretty concerned about this approach. It is weak on policy application, because it mainly favors the eviction of large-size transactions, and not transactions that are large *relative* to the fees they pay.

The single-pair fee comparison is an extremely weak and non-deterministic way of awarding those who offer higher fees a greater chance of relay and confirmation.  We must do better because this is the primary job of the bitcoin network, and the most fundamental service that is offered in exchange for bitcoin.

Remember, randomized eviction is just an implementation cheat - not a policy.  With XT's emphasis on the importance of zero-conf, introducing mempool inconsistency across the network is not good at all.

I have an alternative which uses boost multi-index to apply a deterministic policy that is defined in the index comparator function, without the weird complexities in the core solution.  I'll have it submitted within 24 hours for your consideration.

zander:
@dgenr8 can you please post your objection to the idea (i.e. its not a code-review) to the mailinglist? Discussing concepts on a merge requests doesn't make sense to me.

Also, we have been discussing the approach for weeks now, with various patches being send to the ML in the mean time. Why didn't you voice your ideas then?

dgenr8:
@zander I'm sorry, as usual I thought I would be done much sooner.  Refactors make cherry-picking a real pain in the arse.

mikehearn:
Yep, let's take the design discussions to the list and keep this thread for code review.

@dgenr8 Could you re-post your concerns to the list to start a thread? Copy/paste is fine.

jtoomim:
@dgenr8: I posted a reply to your comments on the bitcoin-xt list: https://groups.google.com/d/msg/bitcoin-xt/9_YHacVC-b4/VtRI0obBEAAJ

jtoomim:
Okay, @gavinandresen candidate comparison debugging lines removed. @zander Timing code removed. @mikehearn Command-line argument is for in-memory bytes (using 2.75x vs serialized bytes as a magic number).

rromanchuk:
I am testing builds of this branch (before the removal of debugging) on two of my nodes (nix and mac os), It's been running smoothly for 24 hours with a reasonable production default  of `maxmempoolbytes=400411697`   Tomorrow I will be switching one of them to non reasonable defaults in  a production environment and will watch the mempool tail closely

mikehearn:
Thanks for the testing, @rromanchuk

rromanchuk:
Hey guys, i've been struggling to get the branches of code i need to run in order to successfully test this. I just took a quick peak at the diff and realized `maxmempooltx` is tightly coupled with `maxmempoolbytes` so it might be helpful to clarify in the command help copy. Looks like using an unreasonable `maxmempoolbytes` setting, of let's say 10mb, the eviction code may never run if `maxmempooltx` is set to something like 50000, or may not run as anticipated by the user.

Just started putting this to work by temporarily squeezing `maxmempooltx` downwards.

jtoomim:
@zander and @dgenr8 noted that the empirical factor borked the default value. It's unborked in 8d51bcc. 8d51bcc tested with default maxmempoolbytes value after a few hours:

```
$ for i in `seq 1 5`; do echo `./bitcoin-cli getmempoolinfo`; sleep 4; done
{ "size" : 5665, "bytes" : 49988347 }
{ "size" : 5676, "bytes" : 49994836 }
{ "size" : 5684, "bytes" : 49997734 }
{ "size" : 5690, "bytes" : 49985555 }
{ "size" : 5697, "bytes" : 49987535 }
```

rromanchuk:
Update: I have been running this on two different production nodes without any problems and working as expected. I also tested this with extreme/nonsensical settings on both test and mainnet without any issues to report. Big fat Tested ACK :checkered_flag: from me.

dgenr8:
The size-weighted limiting has a huge problem: it doesn't stop relaying when the mempool is full.  It will take a spam attack in full swing, and relay every single transaction to all peers, possibly hundreds of them.

This is a direct result of the fact that it doesn't implement a way of prioritizing transactions.  It attempts to treat them all equally in relaying to miners, even though miners are acknowledged to prioritize by feerate.

rromanchuk:
> The size-weighted limiting has a huge problem: it doesn't stop relaying when the mempool is full. 

Well, not if you're using `maxmempooltxns` 

I'll take this to the discussion group, but as a user, I much rather relay spam and evict good actors during a "spam attack" than having my node crash. Node operators shouldn't have to constantly tweak minrelayfees just to keep their nodes running.

jtoomim:
@dgenr8 : I think that regulating the size of a node's mempool and regulation of the network's bandwidth are very separate things, and should operate under different policies. A node can run out of memory long before bandwidth, or vice versa. Miners, for example, currently want to run with a mempool of about 4 MB due to limitations in getblocktemplate, but they shouldn't stop relaying transactions when they get to that level.

jtoomim:
I've mined at least one block with this code on mainnet.

dgenr8:
@jtoomim If we go this way, do you currently know of a reason not to merge this into a release for general testing?

jtoomim:
I know of no obstacles. As I've said, I've used this on mainnet for months.

rromanchuk:
@dgenr8 i've been using this since day 1. Very well tested against edge cases and tested against spam attacks during the testnet bip101 tests.  Probably not as much on everyone's mind since the last spam attack.

dgenr8:
Closing in favor of #90.

PR #91, opened 2015-11-03
Low bandwidth block relay using thin blocks

mikehearn:
Thin blocks are blocks sent between full nodes using a full-match Bloom filter: in other words, as a list of hashes (plus some merkle tree control bits).

In my own tests, the feature makes little difference, but that's because all my internet connections are pretty fast. This work may benefit people behind slow or flaky connections (i.e. Great Firewall) a lot more.

Timings can be seen using -debug=relayperf which now logs the gap between requesting a block and accepting it.

WARNING: This work is only lightly tested.

WARNING: This patch is incomplete. As Core is making Bloom filtering optional, this code needs to check for cases where the remote peer has it disabled.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -198,6 +198,10 @@ bool IsStealthMode() {
     return GetBoolArg("-stealth-mode", false);
 }
 
+bool UsingThinBlocks() {
+    return GetBoolArg("-use-thin-blocks", true);
+}
+
 //////////////////////////////////////////////////////////////////////////////
 //
 // Registration of network node signals.
@@ -1891,6 +1895,8 @@ void static FlushBlockFile(bool fFinalize = false)
 }
 
 bool FindUndoPos(CValidationState &state, int nFile, CDiskBlockPos &pos, unsigned int nAddSize);
+
+void HandleBlockMessage(CNode *pfrom, const string &strCommand, CBlock &block, const CInv &inv);
 
 static CCheckQueue<CScriptCheck> scriptcheckqueue(128);
 
@@ -4420,6 +4426,11 @@ bool ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, int64_t
         int64_t nTimeOffset = nTime - GetTime();
         pfrom->nTimeOffset = nTimeOffset;
         AddTimeData(pfrom->addr, nTimeOffset);
+
+        // Enable thin blocks if we're not doing bulk downloads (it's faster to use ordinary block messages when
+        // catching up with the block chain).
+        if (!IsInitialBlockDownload() && UsingThinBlocks())
+            pfrom->PushMessage("filterload", CBloomFilter());
     }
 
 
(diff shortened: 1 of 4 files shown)


jtoomim:
I think I'll ask people to help test this on testnet with big blocks in a few days. Want to do some experimentation with stock 0.11.0ABCD first.

rromanchuk:
I need to do more investigating, but i noticed when i tried to sync on testnet using this branch, it started crawwwwwwwling :turtle:  When i relaunched with current HEAD it was back to :racehorse: 

Started seeing a lot of these in the log
```
2015-11-08 16:54:38 Peer=25 is stalling block download, disconnecting
2015-11-08 16:54:40 Peer=26 is stalling block download, disconnecting
2015-11-08 16:54:42 Peer=27 is stalling block download, disconnecting
2015-11-08 16:54:44 Peer=28 is stalling block download, disconnecting
2015-11-08 16:54:46 Peer=29 is stalling block download, disconnecting
```

And these
```
2015-11-08 18:51:49 received thin block 0000000000bc915505318327aa0f18568ce024702a024d7c4a3ecfe80a893d6c peer=477
2015-11-08 18:51:57 received thin block 0000000000bc915505318327aa0f18568ce024702a024d7c4a3ecfe80a893d6c peer=533
2015-11-08 18:52:00 received thin block 0000000000bc915505318327aa0f18568ce024702a024d7c4a3ecfe80a893d6c peer=534
2015-11-08 18:52:12 received thin block 0000000000bc915505318327aa0f18568ce024702a024d7c4a3ecfe80a893d6c peer=535
```

mikehearn:
Testnet is pretty borked. I don't normally try things with it anymore as you really have no idea what's going on there at any given time, so reproducing issues can be a nightmare.

It looks from the high peer id count that you've got either tons of peer connections or there's massive churn? And you're downloading the same block from multiple peers, which doesn't seem right.

Anyway, I'm pretty busy this week. If you want to investigate, please do.

jtoomim:
It seems someone is spamming lots of low-diff v4 blocks. This started recently as far as I can tell. I'm thinking of switching to a new genesis block.

rromanchuk:
@mikehearn this will be pretty easy to validate, I will try again from a clean, "out of the box"  sync on mainnet.

jtoomim:
Potential bug: https://gist.github.com/jtoomim/4ac03b7713c4d1099337

jtoomim:
@mikehearn have you had a chance to look at the Exception 'CDataStream::read(): end of data' error? We stopped using thin blocks in our testnet testing because they were too buggy.

mikehearn:
No. I will be busy for a while and don't plan to work on this further. It's here if anyone wants to pick it up and run with it.

jtoomim:
I think what I'm going to try to do is patch this to catch the error which I saw earlier and disable thin block propagation for that block. This will allow thin blocks to work about 99% of the time, and when it fails it will only be at most 2x as slow as defaulting to full blocks.

jtoomim:
Do you think that this might be failing due to false positives in the bloom filter? I.e. block sender thinking there's a particular tx in the recipient's mempool when there isn't actually, due to collisions in the bloom filter's hash space?

mikehearn:
If a transaction isn't pushed by the remote peer then the code asks for it again (and this is visible in the logs when it happens).

jtoomim:
We've got transaction stuff trimmed out of our debug=net logs. Tx stuff is in debug=netfull now in our branch. We probably wouldn't be able to find it without that enabled.

I did this on my testing branch for now:

https://github.com/jtoomim/bitcoinxt/commit/f2940820ba8e3bdd8041ce10463ef4dcd8043ddf

Total hack. Hoping to get some performance data to help figure out if thin blocks have enough of a performance benefit to be worth debugging.

jtoomim:
In case someone is thinking of making a version that includes this code as-is:

This code is buggy. Tested NACK.

dgenr8:
Closing, work continued in #109.

PR #95, opened 2015-11-15
Cherry-pick BIP65 (CHECKLOCKTIMEVERIFY) support

dgenr8:
Cherry-pick CLTV support from core 0.11 backport bitcoin/bitcoin#6707.

XT will join core in counting BIP101 blocks as votes for BIP65, since the BIP101 version 0x20000007 >= 4.

The BIP101 version number is not incremented.


mikehearn:
Ah, I didn't see this before opening up #96 - that one has the low-s change too (which I know you dislike, me too, but it doesn't help to fight that one).

dgenr8:
I fully agree with low-s standardness.

jtoomim:
Merged this into my fortestnet branch, https://github.com/jtoomim/bitcoinxt/pull/1. I'll start deploying it soon. We might be done with the repeated forking, so it might be too late for us to test out CLTV with Core without setting up a node that's connected to Core nodes and not any other XT nodes. That could be arranged, but I probably won't be taking the time to do it myself.

Issue #93, opened 2015-11-10
Supporting BIP65


jl2012:
Adoption of BIP65 has been started and around 1/5 of miners are now mining v4 blocks. XT should merge BIP65 as soon as possible

gavinandresen:
0.11 backport from Core is here:
  https://github.com/bitcoin/bitcoin/pull/6707

... should be easy to merge into XT.

jtoomim:
@gavinandresen @mikehearn I consider this a high priority. If nobody else steps up to merge this (maybe @dgenr8 ? I know he likes pulling stuff from Core), let me know and I can set aside testnet3 stuff to get this done.

zander:
What is the expected result on voting on BIP101 at the same time as voting on BIP65?

Don't they have different, incompatible, voting mechanisms?

gavinandresen:
ACK on supporting BIP65; I'm mostly out until Thursday with meetings, so can't help until then.

Re: voting: will be fine after BIP65 support is merged, assuming all BIP101 miners are also willing to vote for BIP65 (which is a very good assumption).

jtoomim:
@zander : As far as BIP65 is concerned, a vote for BIP65 is a block version > 3. This means that XT is already voting for BIP65. Once 75% of blocks are mined by BIP101 or BIP65 nodes, the BIP65 nodes will enforce BIP65 rules on any node with a version > 3. This means that an invalid OP_CLTV would cause XT blocks to be orphaned by BIP65 nodes, and this could cause a persistent fork if BIP101 hashpower > BIP65 hashpower.

jameshilliard:
> Once 75% of blocks are mined by BIP101 or BIP65 nodes, the BIP65 nodes will enforce BIP65 rules on any node with a version > 3.

This is an interesting side effect of using version numbers for a hardfork attempt without support from most miners, rather than blocks getting orphaned at 95% for non-upgraded clients they get orphaned at 75%. The signaling method for BIP101 should probably be changed, using coinbase sigs for instance would not cause issues like this.

> this could cause a persistent fork if BIP101 hashpower > BIP65 hashpower.

Luckily this is very unlikely to happen on mainnet.

jtoomim:
[...]

PR #96, opened 2015-11-16
CLTV and low s

mikehearn:


dgenr8:
Might want to pull in newly-backported change bitcoin/bitcoin#7021.

jtoomim:
Merged this into my fortestnet branch, https://github.com/jtoomim/bitcoinxt/pull/1. I'll start deploying it soon. We might be done with the repeated forking, so it might be too late for us to test out CLTV with Core without setting up a node that's connected to Core nodes and not any other XT nodes. That could be arranged, but I probably won't be taking the time to do it myself.

dgenr8:
The BIP65 tests need to be updated for the regtest chainparams changes in BIP101.  Here's a commit 50d09a202cfd6876b99c1ee76d16a13e8fa4bfa7 with the required changes.

aquiter:
Were you not telling the truth when you said on bitcoin-development that there was no consensus for this funds freezing feature?

Why would you adopt this complex, failure prone, BlockStream backdoor with a broken deployment system into BitcoinXT? I don't understand. Are you trying to help them win?

mikehearn:
There was indeed no consensus about the rollout mechanism (soft fork vs hard fork). The opcode itself is quite harmless. 

The nature of Bitcoin is that the majority wins. It does not have anything to do with unanimous consensus as so often suggested or implied (by Blockstream and their friends). So XT has no choice in the matter. Whilst miners support Core it has to accept changes to the protocol even if they are badly thought out.

mikehearn:
OK, this seems OK now. @gavinandresen ok with you?

chaosgrid:
Could you comment on this: 

>Wouldn't it be better to bump the version for BIP(101+65) to have clearly defined behaviour?

See https://github.com/bitcoinxt/bitcoinxt/issues/93

gmaxwell:
> Whilst miners support Core it has to accept changes to the protocol

It's my understanding that in your preferred vision of the world that users get forced software update; but that isn't the world of Bitcoin Core.  CLTV is in a freestanding update that does effectively nothing else (there is a windows specific database corruption bugfix; but that is presumably irrelevant to miners).

rromanchuk:
@gmaxwell you know what he means. He doesn't literally think the  blockstream/federal reserve/putin/21co are conspiring for CLTV. He was just responding to a "the sky is falling" comment above him. 


> It's my understanding that in your preferred vision of the world that users get forced software updates; but that isn't the world of Bitcoin Core. CLTV is in a freestanding update that does effectively nothing else (there is a windows specific database corruption bugfix, and some other minor twiddling; but that is presumably irrelevant to miners).

Whoah, that's a lot to unpack. :fearful:

jtoomim:
I ran into an issue in my testnet testing that I think is related to this PR. I don't really know what happened, but here's my leading hypothesis:

1. XT forked off of Core. Core was broadcasting OP_CLTVs (both valid and invalid), and XT included them on its fork.

2. I upgraded some of my XT nodes to include BIP65 support, making some of the transactions in the chain invalid.

3. The Core branch's PoW exceeded XT's, triggering a reorg.

4 (unclear): Some of my XT nodes may have had an OOM event causing a crash, or they may have crashed for another reason, or they may have not crashed at all until the bug in question occurred.

5. During reorg and/or during startup 288 block verification, an invalid CLTV transaction is discovered inside a block. Somehow, this causes the first line of DisconnectBlock() to fail:

    assert(pindex->GetBlockHash() == view.GetBestBlock());

Weird things: sipa says that reorgs shouldn't check signatures, so it shouldn't notice an invalid CLTV, but he thinks that startup validation might find a difference. Looking at the log files, this error seems to happen during reorg, not during startup verification.

I downgraded all of the nodes that were encountering this issue to 0.11.0C (no BIP65 support), and the issue went away. I no longer have any nodes that are encountering this issue on testnet, unfortunately.

mikehearn:
@gmaxwell You know perfectly well what I mean - no, supporting CLTV is now not optional no matter how many times you repeat that utterly ridiculous claim. It is not acceptable that a miner builds on top of an invalid block and produces something that will be rejected - avoiding that is the *entire point* of running a full node, and so XT has to merge this patch set, regardless of whether we think it's a good idea or not.

The fact that you continue to insist that miners can pick'n'mix what they check just goes to show how badly you've lost the plot. It isn't like that.

gmaxwell:
Optional for you? who knows. If a tree falls in a forest that no one uses does it make a sound? To the extent it is, it's not because Core forced you... to the extent that it becomes mandatory for you it will be because the deployed network did so.

You made a statement about miners being forced, and yet pose no mechanism of action for that to be possible.  As I pointed out: You're the guy with the forced automatic upgrade aspirations, not Core.  Miners could chose to not take this feature, -- and you sure tried to submarine it at the 11th hour. It was not tied to anything else. 

Perhaps you'd have more success if you thought less in terms of plots.

jtoomim:
https://gist.github.com/jtoomim/06f6054eafeae08f8717

There's some voluminous extracts from the logfile of one machine that had the problem described 22 hours ago. Here's one of the money sections:

```
2015-11-17 21:35:56.509442 UpdateTip: new best=000000000016b2f5e0e30da6fda9b798cf8dbded94f3457b3113678b20eda9e9  height=605145  log2_work=66.751811  tx=7208708  date=2015-11-16 14:48:52 progress=0.999947  cache=44.7MiB(109697tx)
2015-11-17 21:35:59.765015 - Disconnect block: 1705.08ms
2015-11-17 21:35:59.845904 ERROR: CScriptCheck(): 990ce5372f9113d88b9038cc138e75e2c959d382790ec2d3409746a9d184e0ee:0 VerifySignature failed: Non-canonical signature: S value is unnecessarily high
2015-11-17 21:35:59.846809 ERROR: AcceptToMemoryPool: ConnectInputs failed 990ce5372f9113d88b9038cc138e75e2c959d382790ec2d3409746a9d184e0ee
2015-11-17 21:36:04.200315 send version message: version 70010, blocks=605145, us=139.162.2.241:18333, peer=11
2015-11-17 21:36:04.200500 sending: version (106 bytes) peer=11
2015-11-17 21:36:04.200911 Added connection peer=14 
2015-11-17 21:36:04.201544 Added connection peer=15 
2015-11-17 21:36:04.201735 socket no message in first 60 seconds, 1 0 from 13
2015-11-17 21:36:04.201843 disconnecting peer=13
2015-11-17 21:36:04.202461 Added connection peer=16 
2015-11-17 21:36:04.202714 

************************
EXCEPTION: St9bad_alloc       
std::bad_alloc       
bitcoin in ProcessMessages()       

2015-11-17 21:36:04.203492 ProcessMessages(block, 748705 bytes) FAILED peer=3
```

There's one instance of an St9bad_alloc without a previous `Non-canonical signature` message, but most of bad_allocs come immediately after a high-S. 

I remember seeing an assert failed message noting main.cpp:1825 (the first line of DisconnectBlock in my fortestnet branch), but I can't find that in my logfiles. It may have only gotten dumped to stderr and not to debug.log.

mikehearn:
std::bad_alloc means you ran out of memory. Is this testing with glibc malloc or jemalloc?

jtoomim:
It looked like I had about 1.5 of 4 GB of physical RAM used, and about 4 GB of swap unused at that general time. I was not looking at the moment it crashed, though. 

These machines should all be using glibc malloc.

mikehearn:
I suspect if the node was doing tons of re-orgs or otherwise dealing with the mess on testnet then it fragmented its heap and puked. Perhaps the next release should have both CLTV+low-s+jemalloc? How big would a patch be to make it use that malloc, any idea?

jtoomim:
My suspicion is that there's an assert error in DisconnectTip that is causing the parent thread to go into shutdown mode and that causes any malloc attempts by child threads to return bad_malloc. I have not seen these nodes using a lot of RAM.

Another error, probably the same, which happened during a very small reorg:

https://gist.github.com/jtoomim/0b558a65a7702c250d0c

jtoomim:
A log from another of my servers (which was not mining) which crashed on a reorg triggered by the same block:

https://gist.github.com/jtoomim/e537f4dd6db51e7d74a4

It's odd that I had 6 servers running the same version, and only two crashed on this block.

The block that triggered the reorg was 000000009ed9704cc64dd32c2a8061eba0f507c79cd56944ed481f42541e0d7f. I don't know which block it was that was being disconnected.

mikehearn:
The Bitcoin code is probably not 100% exception safe, so once you take an exception I'd expect the node to be in a partly messed up state after that especially if it was a malloc error.

@gavinandresen Any ideas?

jtoomim:
Clarifying: I think the malloc error is second. I think the assert failure is first, but the assert failure seems to not show up in debug.log.

dgenr8:
During a reorg, an ATMP allocation error occurs in DisconnectTip, and would lead to UpdateTip never being called.  This is a totally corrupt situation which would cause an assert failure on a future call to DisconnectBlock, if not some earlier problem.  So I'm satisfied there is no problem outside of the out-of-memory condition.

@jtoomim are you satisfied that you understand how you ran out of memory?  Were you not running with a limited mempool?

jtoomim:
ATMP?

No, I am not satisfied that I've been getting OOM. I was running with a limited mempool (400 MB estimated real usage or less), and to wit I've never seen any of these nodes show more than about 40% of their physical RAM in use. All of the nodes have at least 4 GiB of physical RAM, and one of the machines that ran into the issue has 14 GiB. As I said earlier, I think it is more plausible that the assert failure was first, and that the std::bad_alloc happens because a different thread tries to make an alloc after the unhandled exception and it is disallowed.

mikehearn:
Why would making an allocation after an exception be disallowed?

Are you running a 64 bit or 32 bit build?

gavinandresen:
Sounds like it is not out-of-memory, but a violation of the implicit assumption that the blockchain your node saved to disk is a valid chain.

The assumption got violated because the chain was saved with one version of the code (that understood OP_CLTV to be a no-op) but then re-used (via a long re-org) with a newer version.

This might be a "won't fix" bug -- the effort required to create a reproducible test case is pretty high, and the likelihood of this actually happening on the main network is probably low. Then again, finding out where the code assumes that blocks/transactions on disk are valid under current rules and changing that code to be more robust in case the validation rules change would be spiffy (assuming it doesn't have a performance impact).

mikehearn:
That sounds very plausible Gavin. If true, it also shows the dangers of this redefine-a-nop rollout strategy. It complicates the code considerably to have to handle the case where an opcode has changed meaning (vs, was guaranteed to never appear in the chain before).

dgenr8:
ATMP = AcceptToMemoryPool

Reorg validation aside, I'm unclear what proximate cause, other than out of memory, is proposed for the observed std::bad_alloc exception.

PR #98, opened 2015-11-21
Add platform details to comment section of user agent.

dagurval:
This was requested on the mailing list a while ago.

Requires boost 1.55. Feature is disabled if older boost. I don't think the complexity of implementing this without boost is worth it.

If adding the Opt class is overdoing it, I can redo this PR without it and the unittests.

I have so far only compiled on Linux.

Screenshot:
useragent

### DIFF
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -5,8 +5,14 @@
 #include "clientversion.h"
 
 #include "tinyformat.h"
+#include "options.h"
 
 #include <string>
+#include <boost/version.hpp>
+
+#if BOOST_VERSION >= 105500 // Boost 1.55 or newer
+#include <boost/predef.h>
+#endif
 
 /**
  * Name of client reported in the 'version' message. Report the same name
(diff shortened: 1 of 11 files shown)


src/clientversion.cpp
```diff
@@ -83,6 +89,37 @@ static std::string FormatVersion(int nVersion)
         return strprintf("%d.%d.%d.%d", nVersion / 1000000, (nVersion / 10000) % 100, (nVersion / 100) % 100, nVersion % 100);
 }
 
+std::vector<std::string> GetPlatformDetails() {
+    std::vector<std::string> platform;
+#if BOOST_VERSION >= 105500
+
+    // Check OS
+#ifdef BOOST_OS_LINUX
```
dagurval: Should be #if, not #ifdef

mikehearn:
Looks real nice! I had imagined that the platform details would go after a / though, as user-agent parsing code should already know how to separate components by slashes. Otherwise existing software that tries to count market share etc will see them all as unrelated.

dagurval:
Thank you!

BIP14 suggests adding them into comments and I think it´s the correct way to do it. Web browsers do it like this as well.

But it appears as you say that bitnodes is using comments as unique software version identifiers, so I´ll add platform info as a subversion string instead. We can always change it back if they fix their user-agent parsers.

dagurval:
OK, it looks like subversion strings also create unique software versions in market share counting software.

I´m not sure what to do. I think I´ll try to contact the people running bitnodes, xtnodes -- and what other sites are tracking market share?

mikehearn:
Well in that case don't worry about it. It's not that important.

dagurval:
I added a missing "#if BOOST_VERSION >= 105500" in a test case and rebased on top of master

dgenr8:

dagurval:
@dgenr8 I have found no issues with it

PR #99, opened 2015-11-22
Dont claim NODE_NETWORK support in stealth mode unless the node supports it

dagurval:
Pruned node would show NODE_NETWORK service bit when in stealth mode.


dgenr8:
@dagurval You were right.  With the change I pushed for here, #109 plus this change would send NODE_BLOOM in combination with an invalid protocol version.  Your original version of this wouldn't have that problem IIRC.

dagurval:
Nice catch. Easier to maintain a whitelist then a blacklist :-). Back to whitelist it is.

dgenr8:
Needs rebase.

dagurval:
Rebased.

Issue #101, opened 2015-12-02
0.11D bip65-cltv test failure: JSONRPC error: CreateNewBlock(): TestBlockValidity failed


error10:
```
Running 2nd level testscript bip65-cltv.py...
  File "/builddir/build/BUILD/bitcoinxt-0.11D/qa/rpc-tests/test_framework/test_framework.py", line 118, in main
    self.run_test()
  File "/builddir/build/BUILD/bitcoinxt-0.11D/qa/rpc-tests/bip65-cltv.py", line 31, in run_test
    self.nodes[1].generate(10)
  File "/builddir/build/BUILD/bitcoinxt-0.11D/qa/rpc-tests/test_framework/authproxy.py", line 126, in __call__
    raise JSONRPCException(response['error'])
Initializing test directory /tmp/testHaj57g
JSONRPC error: CreateNewBlock(): TestBlockValidity failed
Stopping nodes
Cleaning up
Failed
```

Possibly related to #93

sandakersmann:
Ping @mikehearn and @gavinandresen

dgenr8:
@error10 I can't reproduce when building from v0.11D tag.  What are you running exactly?

error10:
@dgenr8 I'm building from this tarball https://github.com/bitcoinxt/bitcoinxt/archive/v0.11D.tar.gz

I'm building on Fedora 23 x86_64. The following packages were installed:

    dnf upgrade
    dnf install gcc gcc-c++ autoconf automake libtool libdb4-devel libdb4-cxx-devel qt-devel boost-devel openssl-devel libcurl-devel protobuf-devel

The build was done like this:

    ./autogen.sh
    ./configure
    make
    make check
    qa/pull-tester/rpc-tests.sh -extended

dgenr8:
Run the test with --nocleanup and look at debug.log in the created /tmp/... directory.

error10:
@dgenr8 I have node0, node1, node2 and node3. It looks like node0 and node2 started up fairly normally and seem to have nothing untoward. node3 never created a debug.log. And node1's debug.log showed nothing unusual except this:

```
2015-12-03 18:07:22 CreateNewBlock(): total size 1000
2015-12-03 18:07:22 ERROR: ContextualCheckBlockHeader : rejected nVersion=3 block
```

dgenr8:
In the node1 log, what is the height= on the last UpdateTip message?

dgenr8:
[...]

PR #102, opened 2015-12-04
Fix blockversion race condition in two regtests

dgenr8:
On some systems, bip65-cltv.py fails because the default 200 BIP101-version blocks are mined before the test begins, causing it to think BIP65 is already in effect.

@error10 does applying this 7-line change fix issue #101 for you?

### DIFF
--- a/qa/rpc-tests/bip65-cltv.py
+++ b/qa/rpc-tests/bip65-cltv.py
@@ -14,6 +14,13 @@ import shutil
 
 class BIP65Test(BitcoinTestFramework):
 
+    def __init__(self):
+        self.num_nodes = 3
+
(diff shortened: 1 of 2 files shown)


error10:
The test now passes. The output is as follows:

```
Running 2nd level testscript bip65-cltv.py...
Initializing test directory /tmp/testZnPRhG
Stopping nodes
Tests successful
```

dgenr8:
@error10 Yes!  Thank you for submitting.

sandakersmann:
Great work guys :+1:

dgenr8:
Added the same change to forknotify.py, per issue #104.

PR #106, opened 2015-12-06
Fix for busy wait loop in SendShaper

ptschip:
When the sendshaper is on and the outgoing
traffic is pegged to the upper speed limit
then cpu will go to 90% on one core.

The best way I've found to see this is to:

1) get the node running first with no traffic shaping and let it stabilize for about 15 minutes
2) set the outbound traffic shaping way down to about 10 or 15KB but have no inbound traffic shaping
3)  wait for the outbound traffic to be pegged at the upper limit you've set.  Once it's pegged there the cpu will start to race, using up about 90% on one core.

The system must be busy to see this.  ( It's harder on a sunday afternoon to get this problem to appear)

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -658,7 +658,11 @@ void SocketSendData(CNode *pnode)
 
         int amt2Send = min((int)(data.size() - pnode->nSendOffset),
                 sendShaper.available(SEND_SHAPER_MIN_FRAG));
-        if (amt2Send==0) break;
+        if (amt2Send == 0) {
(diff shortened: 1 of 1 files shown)


mikehearn:
Your fix doesn't look like a fix for a race condition but rather a busy-wait loop?

ptschip:
ha, true enough, i'll change the title of the PR to "Busy wait loop in
sendshaper" also changed the commit text.

dgenr8:
@ptschip Do you currently know of a reason not to merge this into a release for general testing?

ptschip:
I think it should be merged.

PR #107, opened 2015-12-07
Fix client version so users don't accidentially blockwithold on p2pool

jameshilliard:
I normally wouldn't make a pull request to this repo but there are already users [bypassing](https://www.reddit.com/r/bitcoinxt/comments/3vrf4e/i_made_a_fork_of_latest_version_of_p2pool_with/) p2pool version checks which will cause a withholding attack on p2pool because you haven't set the client version to one that signals BIP65 support.


jtoomim:
I think the best way to solve this issue is to tell people to not bypass the p2pool version checks and to make p2pool check for version 0.11D once it has been released. Feel free to link to https://www.reddit.com/r/bitcoinxt/comments/3vrf4e/i_made_a_fork_of_latest_version_of_p2pool_with/cxqe64h if you see anyone still trying to do this.

jameshilliard:
@jtoomim p2pool currently only looks at the Client Version field, I see no reason that this should be handled on that side, XT is a fork of core and shouldn't try to break compatibility like this.

jtoomim:
forrestv said he would change p2pool to check for the relevant BitcoinXT version once one was released. As no such version has been released yet, he has not made any changes.

Changing the version on the BitcoinXT release would delay the release. Gavin and Mike would have to restart the gitian build and signature process with a new version number. That takes several hours, and can confuse people with the different branding and the inaccurate version numbers (as BitcoinXT is not actually based on 0.11.2, it just contains one feature that is present in 0.11.2). 

In p2pool, you only need to change one line of code. No compilation is necessary. No release process is necessary. The maintainer has already offered to do this there. This should be done there.

jameshilliard:
@jtoomim the problem is that the version number is the same as 0.11.0 currently so that can't really be used to differentiate

jtoomim:
The version string is 0.11.0D, not 0.11.0. That can be used to differentiate. 

If you're saying that this might be a 2-line problem in python to fix (due to another piece of code throwing out non-numeric components, or something like that), that might be right. However, it's still better fixed in python.

jameshilliard:
@jtoomim the version string p2pool uses looks like this under getinfo, it doesn't appear to have any letters:
"version" : 110200

jtoomim:
That can be fixed easily.

jtoomim:
I'll put in a PR on p2pool once I decide that I want to mine on XT again.

jameshilliard:
@jtoomim I don't really see another way to differentiate using the RPC interface

mikehearn:
Why does p2pool do any version checks at all? It shouldn't be making assumptions that users are running Core.

jameshilliard:
@mikehearn P2pool shares aren't fully validated in the same way bitcoin blocks are so it has to rely on bitcoind providing compliant templates. Shares made by a miner using an outdated node would be effectively blockwitholding to other p2pool miners since any blocks mined by that miner wold be rejected.

jtoomim:
@jameshilliard It might be hard to distinguish via the RPC, I don't know. It's easy to distinguish via the p2p protocol. It's also possible to distinguish with a command-line option.

jameshilliard:
@jtoomim yeah, I don't think p2p protocol and command line will work, it needs to be RPC since that's how p2pool interfaces with core

jtoomim:
p2pool interfaces with bitcoind with both the p2p protocol and the rpc interface.

 I'll fix this in p2pool myself if forrestv doesn't.

jameshilliard:
@jtoomim I thought it only interfaced with p2p for relaying, this would be a far simpler fix on the the bitcoind side regardless

mikehearn:
@jameshilliard that's what the version field in the p2p protocol is for.

jameshilliard:
@mikehearn has that changed in XT between 0.11C and 0.11D?

mikehearn:
Yes, it has the CLTV bit set. There are 0.11D pre-release binaries available on github so you can just go and try it.

It's worth noting that this whole mess is a result of soft forks. If these changes were being rolled out as a hard fork, p2pool could just compare advertised chain height from the local node vs the peers, or better, the node itself would just stop vending work. With soft forks, the node doesn't realise anything is wrong, so you end up with weird version hacks like this.

jameshilliard:
@mikehearn P2pool AFAIK is already designed to be used with altcoins so I don't really think it makes much of an attempt to validate more than a basic version check, it basically trusts the bitcoind node to be accurate like a regular pool would. This has little to do with hard/soft forks since pools will blindly build off of whatever bitcoind gives them. Is there any reason the version number wasn't bumped when the CLTV changes were merged?

mikehearn:
Actually, 0.11D doesn't vote for CLTV in the version bits header in the block. It will validate such blocks though. The version number in the subver field was bumped in 0.11D though.

It doesn't advertise itself as 0.11.2 at the RPC level because we didn't incorporate all the changes. Also: again, software shouldn't be relying on it. P2Pool should just swallow whatever the local node gives it. I realise Forrest is trying to be helpful here but the issue is that soft forks are just inherently dangerous for miners. There isn't any good workaround for that.

jameshilliard:
> it doesn't advertise itself as 0.11.2 at the RPC level because we didn't incorporate all the changes.

The problem is less about specifically what you set it to and more that you aren't setting it to anything unique at all between releases. IMO it's probably easiest to just set it to the core version that you are compatible with.

> software shouldn't be relying on it.

There really needs to be a way to version check through the RPC interface one way or another IMO for startup sanity checks.

> Actually, 0.11D doesn't vote for CLTV in the version bits header in the block.

if it's setting version to anything > v3 then other clients are going to interpret it that way

> soft forks are just inherently dangerous for miners

This is really only an issue for p2pool AFAICT and I think p2pool is mostly blind to hardforks as well(depending on what changes). Centralized pools are easy to upgrade.

gavinandresen:
The getblockchaininfo RPC call is the correct way to judge whether or not any particular soft-fork is known.

e.g. running on testnet right now:

```
gavin$ ./bitcoin-cli -testnet getblockchaininfo
{
    "chain" : "test",
    "blocks" : 607833,
    "headers" : 607833,
    "bestblockhash" : "0000000000098284376808e4aff2d40dcf8bd94495fae36389b247bcbf5503e0",
    "difficulty" : 1.00000000,
    "verificationprogress" : 0.99977160,
    "chainwork" : "00000000000000000000000000000000000000000000000790c1feef1fe648f2",
    "pruned" : false,
    "softforks" : [
        {
            "id" : "bip34",
            "version" : 2,
            "enforce" : {
                "status" : true,
                "found" : 100,
                "required" : 51,
                "window" : 100
            },
            "reject" : {
                "status" : true,
                "found" : 100,
                "required" : 75,
                "window" : 100
            }
        },
        {
            "id" : "bip66",
            "version" : 3,
            "enforce" : {
                "status" : true,
                "found" : 100,
                "required" : 51,
                "window" : 100
            },
            "reject" : {
                "status" : true,
                "found" : 100,
                "required" : 75,
                "window" : 100
            }
        },
        {
            "id" : "bip65",
            "version" : 4,
            "enforce" : {
                "status" : true,
                "found" : 100,
                "required" : 51,
                "window" : 100
            },
            "reject" : {
                "status" : true,
                "found" : 100,
                "required" : 75,
                "window" : 100
            }
        }
    ]
}
```

... the version of XT I'm running knows about bip34, 65 and 66.

Hannott:
https://github.com/Hannott/p2pool/commit/766d7ed690fdb3dbf5c4d0fa51baf27bfadc57b5

I tested this with 11C and 11D, and while a bit ugly in case you are running a version that don't have the 'softforks', it seems to work correctly. Would be nice if someone could take a quick look

Hannott:
@gavinandresen @jameshilliard Care to take a look at the latest version? https://github.com/Hannott/p2pool

It looks for bip65 in getblockchaininfo instead of using the version

gavinandresen:
@Hannott : looks good to me (but I don't run p2pool and don't have time to help test)

jameshilliard:
@Hannott do you want to PR that to the main p2pool repo?

Hannott:
@jameshilliard Hope I did this correctly. Haven't worked much with github yet.

https://github.com/p2pool/p2pool/pull/287

jameshilliard:
@Hannott looks fine to me

mikehearn:
Sounds like p2pool was fixed, so that's great. Thanks Gavin for the suggestion.

PR #69, opened 2015-09-12
implement uacomment config parameter

SoreGums:
which can add comments to user agent as per BIP-0014

limit total length of user agent comments
Reworked-By: Wladimir J. van der Laan <laanwj@gmail.com>

Cherry Picked from 7b79cbd722d35b8113d5136b06d4a8e5fd569fc6
merged into StealthMode by SoreGums <nicholas.orr@zxgen.net>

PR #108, opened 2015-12-14
Implement uacomment (alternative to #69)

dagurval:
This PR is an alternative implementation of #69 

This one is compatible with #98. Also, compared to #69 it does not add a global subversion variable, has a unit test and --help text.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -113,6 +113,7 @@ BITCOIN_CORE_H = \
   net.h \
   netbase.h \
   noui.h \
+  options.h \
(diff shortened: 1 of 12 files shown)


dgenr8:

dagurval:
It should perhaps check for reserved symbols in uacomment and refuse to start bitcoind if users uacomment config line has them. These symbols are reserved according to BIP14:
```
/ : ( )
```

dagurval:
I added check for reserved characters. I know of no reason not to merge this now.

PR #112, opened 2015-12-31
Modify mainnet params for 2/4/8, block version=5

dgenr8:
Makes 2 changes to the BIP101 implementation:
 - Initial post-fork block size limit is 2MB
 - The limit doubles twice, at 2-year intervals

Earliest activation is not changed.  Grace period is not changed.

Block version is set to 5.

There is no BIP that corresponds exactly to this PR.  It is BIP101 with the above changes.

No testnet parameters are changed (testnet still supports 8MB and 10 doublings).

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -60,8 +60,8 @@ public:
         // Then, if miner consensus: 8MB max, doubling every two years
         consensus.nMaxSizePreFork = 1000*1000; // 1MB max pre-fork
         consensus.nSizeDoubleEpoch = 60*60*24*365*2; // two years
-        consensus.nMaxSizeBase = 8*1000*1000; // 8MB
-        consensus.nMaxSizeDoublings = 10;
+        consensus.nMaxSizeBase = 2*1000*1000; // 2MB
+        consensus.nMaxSizeDoublings = 2;
         consensus.nActivateSizeForkMajority = 750; // 75% of hashpower to activate fork
         consensus.nSizeForkGracePeriod = 60*60*24*14; // two week grace period after activation
 
--- a/qa/rpc-tests/bigblocks.py
+++ b/qa/rpc-tests/bigblocks.py
@@ -15,7 +15,7 @@ CACHE_DIR = "cache_bigblock"
 
 # regression test / testnet fork params:
 FORK_TIME = 1438387200
-FORK_BLOCK_VERSION = 0x20000007
+FORK_BLOCK_VERSION = 5
 FORK_GRACE_PERIOD = 60*60*24
 
 class BigBlockTest(BitcoinTestFramework):
--- a/qa/rpc-tests/forknotify.py
+++ b/qa/rpc-tests/forknotify.py
@@ -35,9 +35,9 @@ class ForkNotifyTest(BitcoinTestFramework):
                                 ["-blockversion=211"]))
         connect_nodes(self.nodes[1], 0)
 
-        # Node2 mines block.version=0x20000007 blocks
+        # Node2 mines block.version=5 blocks
         self.nodes.append(start_node(2, self.options.tmpdir,
-                            ["-blockversion=%d"%(0x20000007,)]))
+                            ["-blockversion=%d"%(5,)]))
         connect_nodes(self.nodes[2], 0)
 
         self.is_network_split = False
--- a/src/primitives/block.h
+++ b/src/primitives/block.h
@@ -11,7 +11,7 @@
 #include "uint256.h"
 
 /** Blocks with version fields that have these bits set activate the bigger-block fork */
-const unsigned int SIZE_FORK_VERSION = 0x20000007;
+const unsigned int SIZE_FORK_VERSION = 5;
 
 /** Nodes collect new transactions into a block, hash them into a hash tree,
  * and scan through nonce values to make the block's hash satisfy proof-of-work
--- a/src/test/block_size_tests.cpp
+++ b/src/test/block_size_tests.cpp
@@ -18,7 +18,7 @@
 // These must match parameters in chainparams.cpp
 static const uint64_t EARLIEST_FORK_TIME = 1452470400; // 11 Jan 2016
 static const uint32_t MAXSIZE_PREFORK = 1000*1000;
-static const uint32_t MAXSIZE_POSTFORK = 8*1000*1000;
+static const uint32_t MAXSIZE_POSTFORK = 2*1000*1000;
 static const uint64_t SIZE_DOUBLE_EPOCH = 60*60*24*365*2; // two years
 
 BOOST_FIXTURE_TEST_SUITE(block_size_tests, TestingSetup)


jtoomim:
I think that modifying BIP101 to do 2-4-8 is a great idea. I think having XT support this is also a good idea.

F2Pool wants their forks and block sizes to be triggered and keyed by block height, not by timestamp. Can you make that modification too?

It would also be worthwhile to look into different activation conditions, such as a simple flag day option (which I don't like, but others might). Using the current activation mechanism, I think we should tweak the parameters to be more conservative -- I suggest 90% support with a 1 month grace period. I think we are likely to be able to reach that threshold. It might even be best to do 95% support. Don't know. Not attached to any one thing, but I think that 75% is going to cause the cause more harm than good right now.

Some of the pools (e.g. BTCC) are open to using a fork other than Core master. However, they have explicitly named XT as not being an option. I suggest rewriting this as a fork off of Core master. I'd do it myself, but I'm gonna be busy for a week or two.

dgenr8:
Unfortunately, I don't think core will do anything unless miners actually line up behind a fork.

[Thanks to you](https://imgur.com/3fceWVb), we know what miners say they support.  Let's give them a way to prove it.

jtoomim:
> Unfortunately, I don't think core will do anything unless miners actually line up behind a fork.

Some and possibly most of the miners are willing to line up behind a fork, as long as it is a fork of Core master.

> Thanks to you, we know what miners say they support. Let's give them a way to prove it.

That image is out of date. See https://docs.google.com/spreadsheets/d/1Cg9Qo9Vl5PdJYD4EiHnIGMV3G48pWmcWI3NFoKKfIzU/edit#gid=0. Of note, Bitfury revised their position, and BW added theirs. Also, there is a partial translation into Chinese.

berrtus:
Hi guys, just want you to know I am on board here.  I have been working for this option both at the google groups, and reddit.  jtoomim has really done some great work in in development and testing.  Thanks to him we have all the data necessary to argue our point.  In particular we know we best represent the option that the miners and business community can most likely support.  2-4-8 is the most sensible and practical option.  This option takes out the unnecessary forward predictions that we won't be using anyway because we are going to develop dynamic models instead.  The block limit should only be a last resort and way above what is happening in the code which is what should be used to determine the block size.

amacneil:
Sorry if others already know this, but does this implementation have a "smooth" increase in block size over time, or does it literally double at fixed intervals every two years?

I believe I heard Gavin mention that the existing BIP101 code had smooth increases over time, which I believe is a good thing (BIP202 has similar properties).

dgenr8:
@amacneil It increases linearly between the doubling points.

berrtus:
I agree with jtoomim about changing the activation condition but think 80% is sufficient.  I don't know enough about the details of the code to suggest core over xt, but would say: If it was core we have to take out RBF and any connections they have been rigging to support their LN.

src/chainparams.cpp
```diff
@@ -60,8 +60,8 @@ class CMainParams : public CChainParams {
         // Then, if miner consensus: 8MB max, doubling every two years
         consensus.nMaxSizePreFork = 1000*1000; // 1MB max pre-fork
         consensus.nSizeDoubleEpoch = 60*60*24*365*2; // two years
-        consensus.nMaxSizeBase = 8*1000*1000; // 8MB
-        consensus.nMaxSizeDoublings = 10;
+        consensus.nMaxSizeBase = 2*1000*1000; // 2MB
+        consensus.nMaxSizeDoublings = 2;
```
dgmib: I suggest leaving this at nMaxSizeDoublings = 10;

It wouldn't make a difference until sometime after 2020, and prevents us from being in same place in the same situation again in 4 year where we need a hard fork to keep growing.

jtoomim: Bitfury would object to and likely block such a change. A few other pools also might.

berrtus:
>I suggest leaving this at nMaxSizeDoublings = 10;

The whole point is we are getting away from forward predictions that are too far out.  The odds of these guesses being successful tends to zero quicker than we think.  They will be unnecessary as we will have dynamic code, even Gavin Andresen has referred to it.  And finally for all that zero gain we put off most of the hashing power.  Yes, by some super small chance the guess could turn out to be correct and used, but making the guess itself decreases the odds by much more that the code would ever be considered for use.

dgmib:
I 100% agree with you we can't make forward predictions that far out.  Heck, we can't even make accurate forward predictions a few weeks out let alone a few years.  I don't expect to be able to guess this correctly at all. But no matter what you do about the block size, you're making a guess today.  Making the limit stop growing at 8MB is making a guess today that 8MB will be enough.

Which is exactly why I'm arguing it should keep doubling, we can't possible know today when to stop the block size from growing.  We can't possible know if 8MB is the best limit or 16MB or any other number.  The block size limit should stop growing when we have a reason and make the choice to stop it from growing.

I'm advocating that we adopt a strategy of "future inaction won't change the economic conditions".  I'm just fine with changing the economic conditions someday, I just want to see that happen due to deliberate action of the community.

All that said, if Jon is correct, and leaving it a 2 doublings is required to get the miners on board.  So be it.

berrtus:
>All that said, if Jon is correct, and leaving it a 2 doublings is required to get the miners on board. So be it.

Quick recap:  
* The chance we guess correctly tends to zero quickly.
* As we go out to 2020 we almost surely will have dynamic code instead.
* By guessing we open ourselves up to criticism and put off miners thereby reducing not increasing the chance that the code runs to begin with which is more important than the many year out guess.
* The whole point of 2-4-8 is it just goes to 8 otherwise it's really back to being xt.  (I know xt is starting at 8mb)
* There will be likely other reasons for hard forks before 2020 including a properly done SegWit.  Assuming we want that at all.
Additionally:
* I would add hard forks have been made out to be something negative when in fact it's just a way of continuing to get community consensus.  We are not Core trying to get our way with passive aggressive soft forks that lack consensus.  Nor are we trying to get our way by advance for so many years out.  We want continued consensus and are not afraid of it.

berrtus:
We need the following before we can get listed at Coin Dance or xtnodes.com:
Homepage Link  Not same as Github / Download Page Link / FAQ/Help Page Link  // Downloadable files
I have a hosting package and could make the homepage link and format for that page, but would prefer if someone else was helping write it.  Are we all okay with a fork of xt or do we want to fork Core?

jtoomim:
I do not mind having a version of this built off of the XT code. However, I do object to not having a version of this built off of the Core code. The miners have been very explicit that they want this in Core, and I intend to follow their wishes. 

Besides, Core has a lot of good code that's been added recently. I've wanted to use that code a lot while hacking on XT stuff. I disagree with some of the changes that they've made, but most of the changes have been pure technical improvements, and not politically charged crap.

If you guys are focusing on an XT-based branch, I will not be investing much effort in it, and will instead be working on a branch that I'll make off of Core. I may try to cherry-pick code into that new Core fork.

toomim:
@barmstrong (just blogged a cool idea)[https://medium.com/@barmstrong/bitcoin-s-elegant-upgrade-mechanism-miner-voting-66faa35d27af#.au9xqnlgc] for how to count votes:

> I’d like to propose a new way to think about measuring votes where only explicit votes are counted. Right now, default blocks are being counted as a vote for bitcoin core, but it could be that miners haven’t had a chance to upgrade or they don’t want to vote at all (as they’ve stated).

> We should give miners the option of making an explicit vote for bitcoin core if they want it, or a way to abstain and not vote at all.

It would be very helpful to see if miners are actually voting *against* this proposal, or just are too lazy to consider it. Is there a way to let them vote and say that they actively *reject* it, vs. just not bothering to vote?

luke-jr:
Some miners know they have no say in hardforks, and will refuse to vote on such things period on principle.

jtoomim:
@toomim I can add that pretty easily to my spreadsheet... Done. 

https://docs.google.com/spreadsheets/d/1Cg9Qo9Vl5PdJYD4EiHnIGMV3G48pWmcWI3NFoKKfIzU/edit#gid=1056384823

mikehearn:
@jtoomim Most of the new code in Core is crap and would have to be removed anyway (e.g. the RBF stuff, the SegWit soft fork, the connection count limiting etc). 

So I'm not in favour of changing XT to do something that isn't specified as a BIP anywhere, has bizarre changes with no justification (e.g. block height instead of timestamp, why?), and which miners are already saying they won't run (again, without justification).

I know I said a few days ago in email I'd be OK with this, but I changed my mind. I'm going to close this. As 2/4/8 is a bad technical choice anyway (2mb is nowhere near enough excess capacity for a robust network even with current conditions) and the only justification for it is "miners say they might consider that", but miners are also not yet ready to accept that Core is a writeoff, I see no reason to use XT for this.

PR #115, opened 2016-01-17
[walletdb] Add missing LOCK() in Recover() for dummyWallet

dgenr8:
This fix is important to me because the combination of debug settings and -salvagewallet is helpful in testing double-spends.


dagurval:
utACK

PR #119, opened 2016-01-27
Re-enable known-inventory filter in merkleblock

dgenr8:
This is required to serve thin blocks.

We accept the same 1e-6 false-positive rate that is already accepted in the other places the known-inventory filter is used (eg RelayTransaction).

The filter was disabled in core version 0.12 by https://github.com/bitcoin/bitcoin/commit/ec73ef37eccfeda76de55c4ff93ea54d4e69e1ec.


dagurval:
ACK

PR #120, opened 2016-01-29
Thin block fixes

dagurval:
Change the "SupportThinBlocks" criteria to avoid connecting to Bitcoin Core 0.12 nodes. 

Second is to let ThinBlockWorker hold the "is re-requesting transactions" state, rather than CNodeState. This is how it was supposed to be originally, ThinBlockWorker knows when to reset it.


dagurval:
Added a commit to enable use of thin blocks by default.

dgenr8:
With thin blocks enables, I'm getting a loop in ``getchaintips.py`` at line 46, ``self.join_network ()`` where it connects two networks with different tips.

node1 log shows repeated block fetching.

dgenr8:
All tests pass, with and without thin blocks enabled.

PR #121, opened 2016-01-31
Display coinjoins properly in QT

raedah:
Big diff from modifying the main conditional, but actually only a few lines changed. I just got rid of the first "if (nNet > 0 || wtx.IsCoinBase())" statement, fixed the indentation, and moved the "// Credit" section down into an "else if (fAllToMe)" block. Then added a comment where the coinjoins are caught. Thats all.

This fixes the transaction list so that it shows 1 row per transaction instead of 2 for coinjoins, and for the amount shows the transaction Net value. More importantly, this fixes the details view for coinjoin transactions so that Debits and Credits are shown accurately, allowing one to discern what Debits and Credits were theirs in this transaction.

### DIFF
--- a/src/qt/transactiondesc.cpp
+++ b/src/qt/transactiondesc.cpp
@@ -142,13 +142,6 @@ QString TransactionDesc::toHTML(CWallet *wallet, CWalletTx &wtx, TransactionReco
             strHTML += "(" + tr("not accepted") + ")";
         strHTML += "<br>";
     }
-    else if (nNet > 0)
-    {
-        //
-        // Credit
-        //
-        strHTML += "<b>" + tr("Credit") + ":</b> " + BitcoinUnits::formatHtmlWithUnit(unit, nNet) + "<br>";
-    }
     else
     {
         isminetype fAllFromMe = ISMINE_SPENDABLE;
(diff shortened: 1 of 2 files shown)


dgenr8:
Has this been submitted to core?

raedah:
Yes, two and half months ago. No traction. pull request #7101.

raedah:
Coinjoins can be generated using  https://github.com/JoinMarket-Org/joinmarket/

dgenr8:
utACK

dgenr8:
@raedah Could you take a look at [this question](https://groups.google.com/d/msg/bitcoin-xt/FCdTN1oKbkY/uBkkZ3iXHgAJ) that was asked on the mailing list?  The problem described could conceivably be related to this change, although the user did not isolate the problem very convincingly to the release that contains it.

raedah:
bug here in transactionrecord.cpp

I had removed line 45,  `if (nNet > 0 || wtx.IsCoinBase())`
and replaced it with (moved to) line 119,  `else if (fAllToMe)`
But fAllToMe doesn't work because there is likely a change address that is returning to the sender, so all the credits are not ToMe.  So `sub.type = TransactionRecord::RecvWithAddress;`   doesnt get applied.

I can fix it if you want, but I suggest reverting the whole pull, as I am now aware that the console listtransactions needs to be updated as well and am working on a larger patch that will include that.  My current progress is at core pull request 7569. Thanks

dgenr8:
@raedah Thanks, be sure to submit here too when you're ready.

PR #122, opened 2016-02-06
LevelDB fix to reduce corruption after crashes

dgenr8:
Reduces the incidence of corruption requiring reindex on the Windows platform.

This is bitcoin/leveldb#9.


dgenr8:
Results of ``contrib/devtools/git-subtree-check.sh src/leveldb``
```
src/leveldb in HEAD was last updated to upstream commit 20ca81f (tree 1c185cba2ddcb6a77ece1094bc22aea072db84d1)
src/leveldb in HEAD currently refers to tree 1c185cba2ddcb6a77ece1094bc22aea072db84d1
GOOD
```

PR #123, opened 2016-02-07
Support waiting for regular pong and thin block pong at the same time.

dagurval:
Fixes a bug where a thin block pong is interpreted as a regular pong.
This caused the block download to time out.


dgenr8:
utACK

PR #124, opened 2016-02-08
Add pruning to options dialog

dagurval:
I believe it to be a common misconception that running a full node requires > 60 GB of dedicated space. I think the pruning feature should be more available.

This PR adds pruning to options dialog as seen on this screenshot:
screen shot 2016-02-08 at 10 52 57

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -118,6 +118,82 @@
         </layout>
        </item>
        <item>
+        <widget class="QGroupBox" name="groupBox_3">
+         <property name="title">
+          <string>Disk usage</string>
+         </property>
+         <layout class="QVBoxLayout" name="verticalLayout_3">
+          <item>
+           <widget class="QCheckBox" name="pruneEnableChk">
+            <property name="text">
+             <string>Limit blockchain size (prune)</string>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <layout class="QHBoxLayout" name="horizontalLayout">
+            <item>
+             <widget class="QLabel" name="label">
+              <property name="text">
+               <string>Use at most</string>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <widget class="QDoubleSpinBox" name="pruneSize">
+              <property name="suffix">
+               <string> GB</string>
+              </property>
+              <property name="minimum">
+               <double>0.550000000000000</double>
+              </property>
+              <property name="maximum">
+               <double>9999.000000000000000</double>
+              </property>
+              <property name="value">
+               <double>200.000000000000000</double>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <spacer name="horizontalSpacer_3">
+              <property name="orientation">
+               <enum>Qt::Horizontal</enum>
+              </property>
+              <property name="sizeHint" stdset="0">
+               <size>
+                <width>40</width>
+                <height>20</height>
+               </size>
+              </property>
+             </spacer>
+            </item>
+           </layout>
+          </item>
+          <item>
+           <widget class="QLabel" name="diskRequiredLbl">
+            <property name="text">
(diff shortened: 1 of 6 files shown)


dgenr8:
@Diapolo may have useful feedback.

src/qt/forms/optionsdialog.ui
```diff
@@ -118,6 +118,69 @@
...
+               <double>200.000000000000000</double>
+              </property>
+             </widget>
+            </item>
+           </layout>
+          </item>
+          <item>
+           <widget class="QLabel" name="label_2">
+            <property name="text">
+             <string>Note that additional disk space (~2GB) is required for other datasets</string>
```
MarcoFalke: If you cherry-pick https://github.com/bitcoin/bitcoin/commit/b1a8374aaa82e08de8ac9353e76be3cdf43cd3d5, you'd be able to replace the hardcoded value with `CHAIN_STATE_SIZE` (which is the size of the "empty" datadir)

src/qt/optionsmodel.cpp
```diff
@@ -90,6 +90,19 @@ void OptionsModel::Init()
         settings.setValue("nThreadsScriptVerif", DEFAULT_SCRIPTCHECK_THREADS);
     if (!SoftSetArg("-par", settings.value("nThreadsScriptVerif").toString().toStdString()))
         addOverriddenOption("-par");
+    if (!settings.contains("bPrune"))
+        settings.setValue("bPrune", false);
+    if (!settings.contains("nPruneSize"))
+        settings.setValue("nPruneSize", 200.0); // GB
+    if (settings.value("bPrune").toBool()) {
+        int mb = static_cast<int>(settings.value("nPruneSize").toDouble() * 1000);
```
MarcoFalke: Can you check this again. I think `-prune` is one of the options in MiB, so I'd prefer to have consistency between bitcoind and bitcoin-qt.

dagurval:
@MarcoFalke Thank you for the feedback! I made the following changes based on it:
- Cherry picked your intro commit.
- Use CHAIN_STATE_SIZE constant in label text
- Use BLOCK_CHAIN_SIZE as default block limit setting
- Multiply with 1024 instead of 1000

Additional changes
- Use MIN_DISK_SPACE_FOR_BLOCK_FILES for setting minimum allowed prune size
- Added horizontal spacer to make input box smaller

src/qt/guiconstants.h
```diff
@@ -53,4 +53,9 @@ static const int MAX_URI_LENGTH = 255;
 #define QAPP_APP_NAME_DEFAULT "Bitcoin-Qt"
 #define QAPP_APP_NAME_TESTNET "Bitcoin-Qt-testnet"
 
+/* Minimum free space (in GB) needed for data directory when pruned; Does not include prune target */
+static const uint64_t CHAIN_STATE_SIZE = 2;
+/* Minimum free space (in GB) needed for data directory */
+static const uint64_t BLOCK_CHAIN_SIZE = 80;
```
MarcoFalke: I guess it's ok to just move them below `MIN_DISK_SPACE_FOR_BLOCK_FILES` in main.h, so the daemon can use  them as well, just in case.

dagurval: main.h is bloated enough as is, I think we should move them only when we need them

dgenr8:
We don't yet support running a wallet with pruning enabled, so this only benefits people running infrastructure nodes who use the GUI.

src/qt/guiconstants.h
MarcoFalke: Fine with me. You may want to create a pull request against bitcoin/master to get some more review.

MarcoFalke:
* Please use integer (the same as with `-prune`); I have already prepared that and you can cherry-pick https://github.com/MarcoFalke/bitcoin/commit/b21607ed7f2f9baa8863ab12a8b2144112498646.
* Please make sure this cannot be set with `-txindex`; "Prune mode is incompatible with -txindex."

Issue #128, opened 2016-02-14
Update apt repo?


Nutomic:
I'm installed XT from the [apt repo](https://bitcoinxt.software/apt.html). The latest available version is still v0.11.0D-b504e06.

Can you please update that to the latest version?

mikehearn:
The apt repo is no longer maintained, sorry.

Nutomic:
Too bad. I suggest you remove the link from the website then. Also, the download link still points to v0.11.D

dgenr8:
Pinging @stevemyers just in case.

stevemyers:
Hi sorry, just saw your ping.  I'll update the binary versions in the gitian-debian build.sh script and update the README.md file to point to the correct binaries also.  Look for a pull request coming soon.

dagurval:
@Nutomic Can you try to do this and then apt-get update and apt-get upgrade? This repo is signed with the same key as the old one so there should be no warnings.

`
echo 'deb [ arch=amd64 ] http://xt.allnodes.xyz jessie main' > /etc/apt/sources.list.d/bitcoinxt.list
`

Nutomic:
Works perfectly, thanks! Makes running a node so much easier :)

PR #129, opened 2016-02-17
#128 Updated gitian-debian build script and README.md to version 0.11E

stevemyers:
@dgenr8 I tested updated gitian-debian build instructions and upgrade installation on Debian 8 VM.  New 0.11E package installed and started daemon with no errors.


dagurval:
ACK

I tested building and installing.

PR #139, opened 2016-03-26
SPV: Push unconfirmed ancestors to peer. SPV test.

dgenr8:
Lets an SPV peer better track the confirmation process of wallet transactions by knowing details about unconfirmed ancestors.

Connection bloom filter nFlags bit ``BLOOM_ANCESTOR_UPDATE_BIT = 4`` is newly defined to mean that the SPV peer wants its filter be automatically updated with the txids of unconfirmed ancestors of transactions that are already contained in the filter.

Filter update with unconfirmed ancestors occurs in the two places that filtered results for unconfirmed transactions are provided: transaction relay, and the "mempool" p2p response.  The results also include the ancestors added.

SPV "merkleblock" response is indirectly enhanced, because the filter will contain id's for transactions that were at some point seen as unconfirmed ancestors of a wallet transaction.  Double-spend relay alerts are provided when an ancestor is respent, but those relays necessarily exclude malleability clones (which could be produced by a 3rd party).

The included test ``spv.py`` relies on new bloom filter support added to ``mininode.py``.  Additional tests of SPV server functionality can be added in the future.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1364,8 +1364,11 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
                 doubleSpendFilter.clear();
             doubleSpendFilter.insert(relayForOutpoint);
 
-            if (!Opt().IsStealthMode())
-                RelayTransaction(tx);
+            if (!Opt().IsStealthMode()) {
+                std::vector<uint256> vAncestors;
+                vAncestors.push_back(hash); // Alert only for the tx itself
+                RelayTransaction(tx, vAncestors);
+            }
         }
         else
         {
@@ -5050,7 +5053,9 @@ bool ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, int64_t
             if (!AlreadyHave(inv) && AcceptToMemoryPool(mempool, state, tx, true, &fMissingInputs))
             {
                 mempool.check(pcoinsTip);
-                RelayTransaction(tx);
+                std::vector<uint256> vAncestors;
+                mempool.queryAncestors(tx.GetHash(), vAncestors);
+                RelayTransaction(tx, vAncestors);
                 vWorkQueue.push_back(inv.hash);
 
                 LogPrint("mempool", "AcceptToMemoryPool: peer=%d %s: accepted %s (poolsz %u)\n",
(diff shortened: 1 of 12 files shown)


dgenr8:
Some relevant discussion here https://github.com/bitcoin/bitcoin/issues/7237.

schildbach:
Yes, that would be a useful feature!

Is there any garuantee about in what order transactions arrive? (e.g. child before parent)

dgenr8:
@schildbach No guarantees.

src/main.cpp
```diff
@@ -5050,7 +5053,9 @@ bool ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, int64_t
             if (!AlreadyHave(inv) && AcceptToMemoryPool(mempool, state, tx, true, &fMissingInputs))
             {
                 mempool.check(pcoinsTip);
-                RelayTransaction(tx);
+                std::vector<uint256> vAncestors;
+                mempool.queryAncestors(tx.GetHash(), vAncestors);
```
dagurval: Do we need to call queryAncestors if peer has not loaded a filter that wants ancestors? Or does it not matter performance wise?

dagurval:
Great work! utACK

src/main.cpp
dgenr8: Nice catch. Thanks!

schildbach:
@dgenr8 I wonder what would be a good approach as an SPV client. How does an SPV client know all ancestors (in the mempool) have been sent? We can go back to the timeout approach, but it's flaky. Currently all ancestors are getdata'ed and reject messages signal the non-presence of that ancestor. While this is badly performing, it's a more correct approach.

Also, if an ancestor is sent before its parent, how does an SPV client know it's an ancestor (and not a false positive of the filter)? Feels like if using this feature, all received transactions must be kept, just in case…

dgenr8:
@schildbach 
If SPV client's peer has this change, and it is enabled, any unconfirmed ancestor that isn't relayed within seconds after child relay is either known to the client already (having been pushed just before the child, or possibly earlier if it happens to involve the SPV client directly), or will not be available from this peer by getdata.

I don't think I can get you a tidy ancestor package without changing the protocol, at least slightly.

PR #142, opened 2016-04-06
Extend checkpointing to a safe distance in the past

dagurval:
This is a cherry-pick of https://github.com/bitcoinclassic/bitcoinclassic/pull/143

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -59,6 +59,7 @@ BITCOIN_TESTS =\
   test/miner_tests.cpp \
   test/multisig_tests.cpp \
   test/netbase_tests.cpp \
+  test/options_tests.cpp \
   test/p2p_protocol_tests.cpp \
   test/pmt_tests.cpp \
   test/policyestimator_tests.cpp \
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -798,15 +798,6 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (nMempoolSizeLimit < 0 || nMempoolSizeLimit < nMempoolDescendantSizeLimit * 40)
         return InitError(strprintf(_("Error: -maxmempool must be at least %d MB"), GetArg("-limitdescendantsize", DEFAULT_DESCENDANT_SIZE_LIMIT) / 25));
 
-    // -par=0 means autodetect, but nScriptCheckThreads==0 means no concurrency
-    nScriptCheckThreads = GetArg("-par", DEFAULT_SCRIPTCHECK_THREADS);
-    if (nScriptCheckThreads <= 0)
-        nScriptCheckThreads += boost::thread::hardware_concurrency();
-    if (nScriptCheckThreads <= 1)
-        nScriptCheckThreads = 0;
-    else if (nScriptCheckThreads > MAX_SCRIPTCHECK_THREADS)
-        nScriptCheckThreads = MAX_SCRIPTCHECK_THREADS;
-
     fServer = GetBoolArg("-server", false);
 
     // block pruning; get the amount of disk space (in MB) to allot for block & undo files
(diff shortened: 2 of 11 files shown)


dgenr8:
Suppose a script engine / crypto bug causes a network split, or that a way is found to make an obviously invalid script seem valid. 

An upgrade would be issued to fix the bug, but this creates a 24-hour deadline (in the 1 thread case), after which the node cannot heal its chain unless it disables auto-checkpointing, in addition to fixing the bug.

This seems dangerous to me..  At least it seems 24 hours is too aggressive.

christophebiocca:
While I agree with your assessment, this patch actually sets 24hrs * number of validation threads as its threshold.

dagurval:
Would you be fine with this feature if the time barrier was increased @dgenr8? For example from 1 day to 1 week * n validation threads?

christophebiocca:
At that point you're better off just using recent checkpoints/allowing the user to specify a checkpoint.

gavinandresen:
@dgenr8 : If a node accepts an invalid block as valid, it will get the right chain as soon as the rest of the network builds a longer chain.

If you are imagining a network split with 51% or more of hashpower on the wrong side of the split for more than a day...

A. That is extremely unlikely
B. There is a built-in warning if two high-work forks of the chain are detected.
C. If it did happen, the miners who are accepting the bad chain can simply upgrade and then either use the 'invalidateblock' RPC call or start-up with checkpoints disabled to get onto the correct chain.

christophebiocca:
> start-up with checkpoints disabled to get onto the correct chain.

I don't think that'd work, since the invalid block is already "validated". I haven't seen any bookkeeping of which blocks were really validated vs. had signature checks skipped.

dgenr8:
Maybe hypothetical (since this codebase doesn't have a large % of nodes or hashpower) but shouldn't we avoid relying on "the rest of the network?"

A dynamic trailing no-scriptcheck-point seems superior to compiled-in checkpoints but I think the checkpoint age should be on the order of the time it takes the network to recognize a widespread obvious problem, issue a well tested software update and get a supermajority deployment.  Writing this I think 30 days would be better.

With the age dependent on # of threads, it becomes unknown how far back the network as a whole will stop checking scripts. Maybe I'm being over-cautious but I think 30 days would still be a big improvement over fixed checkpoints.

gavinandresen:
Different defaults for Classic and XT (and Core and btcd and...) is a great idea; diversity helps robustness.

So ACK from me on adding 30 days (although isn't a week plenty of time to notice and react? Prior emergencies were resolved in less than 24 hours...).

dagurval:
I agree both that you are over-cautious **and** that 30 days is an improvement over fixed checkpoints :smile: 

I set the default to be 30 * script threads. How about allowing this value to be configured? I added two more commits:
* Moved nScriptThreads from main.h (so that I didn't have to include main.h) + tests
* Added option 'checkpoint-days' + tests

Users can set 'checkpoint-days' to 1 to have the same behaviour as classic.

dgenr8:
At compile time:
```
qt/optionsdialog.cpp: In constructor ‘OptionsDialog::OptionsDialog(QWidget*, bool)’:
qt/optionsdialog.cpp:48:40: error: ‘MAX_SCRIPTCHECK_THREADS’ was not declared in this scope
     ui->threadsScriptVerif->setMaximum(MAX_SCRIPTCHECK_THREADS);
                                        ^
qt/optionsmodel.cpp: In member function ‘void OptionsModel::Init()’:
qt/optionsmodel.cpp:90:50: error: ‘DEFAULT_SCRIPTCHECK_THREADS’ was not declared in this scope
         settings.setValue("nThreadsScriptVerif", DEFAULT_SCRIPTCHECK_THREADS);
```

dagurval:
Oh, I didn't notice I was compiling without Qt.

I replaced a couple of _main.h_ includes with _options.h_ in the Qt code and squashed it into the nScriptCheckThreads commit. bitcoin-qt compiles now.

dgenr8:
ACK

dgenr8:
I worked with a statistics expert to calculate the likelihood of evil miners colluding to push blocktime back by 30 days, to make nodes with this feature accept invalid blocks:

control-mtp


With only 20% honest hashpower, it's virtually guaranteed that blocktime will be reset to a sane value by the honest minority within 4230 blocks, erasing any effort to restrict blocktime to minimal increases and push it into the past.

gmaxwell:
You should fire your "statistics expert", that analysis is wrong on its face-- at all points for honest hashing below 0.5 the probability should be zero.  I pointed this out to you on IRC a week ago-- I'm disappointed that you left this misinformation up without correction.

zander:
@gmaxwell actually, from your comment it looks like you don't seem to understand the actual statement of the statistics expert. Your comment has zero relation to the actual statement made.

It helps if you fully read the actual post fully you go and tell them they are wrong ;)

Specifically;
> at all points for honest hashing below 0.5 the probability should be zero

This was addressed in a different usecase than the one you attack;
> Of course, 50% or more could resort to a DoS attack instead.

I suggest re-reading the thing you attack.  This is the crucial sentence;

> The honest miners have 30 days to get 6 out of any 11 consecutive blocks.

PR #144, opened 2016-04-18
Fix typo that broke the build

mrkline:


dagurval:
I forgot _--with-gui_ again. Thanks.

ACK

dgenr8:
... and I forgot to rebuild after @dagurval fixed the first GUI bug.  Shame!

PR #145, opened 2016-04-18
Fix class/struct mismatches

mrkline:
Several types were declared as structs (presumably to save a `public:` line), but were forward-declared throughout the code as classes.

I changed them back to `class` since it seemed like the path of least resistance (otherwise all the forward declarations would have to be changed).


dagurval:
gcc does not warn/care about this so I didn't notice. Thanks.

utACK

mrkline:
Ah! I'm building with clang - that would explain it.

dgenr8:
utACK

PR #147, opened 2016-04-30
Corrections to bad-chain alert triggering

dgenr8:
Two corrections are made to the bad-chain alert triggering mechanism: oversampling is eliminated and the test is changed to make correct use of the poisson CDF.

Overall, the effect will be for the alert to trigger less often, since it was triggering too easily before.

There has been review and discussion at https://github.com/bitcoin/bitcoin/pull/7568.  The holdup there seems to be that this doesn't fix everything.  No disagreement that it's a big improvement.


dgenr8:
The core project appears to be ready to completely disable all detection of unusually low or high block generation rate, see https://github.com/bitcoin/bitcoin/pull/8275.

Issue #131, opened 2016-02-25
[apt] Move ntp to recommends


dagurval:
Requested by [/u/RussianNeuroMancer)](https://www.reddit.com/r/bitcoinxt/comments/4704sd/aptrepository_has_been_updated_new_url/d0cxdha).

Currently the BitcoinXT apt package depends on installing ntp. Is there a good reason for this or can we make it optional (recommended)?

RussianNeuroMancer:
If systemd-timesyncd is installed and running (timedatectl status) then as I understand synchronization can be configured via /etc/systemd/timesyncd.conf or via NTP= option in network interface description for networkd (again, if enabled and used).

My point is that there is other options for time synchronization besides ntpd.

RussianNeuroMancer:
Seems like latest package still depend on ntp daemon.

dgenr8:
@RussianNeuroMancer Are you able to submit a PR?

RussianNeuroMancer:
I am not developer, I can't write fix for that.

dagurval:
@RussianNeuroMancer Can you try out the .deb package in https://github.com/bitcoinxt/bitcoinxt/issues/154 (first comment)? Does it solve this issue?

RussianNeuroMancer:
Yes, with this deb package ntp can be removed now. Thanks!

RussianNeuroMancer:
Seems like deb package in repository was built with old build.sh, so this change doesn't included in Debian repository.

dagurval:
There has not been a new release since the fix was merged.

It will be included in the next release.

RussianNeuroMancer:
Ok, I see. Thanks for fast answer!

dagurval:
Thank you for running XT :-)

PR #154, opened 2016-06-30
Move ntp from Depends to Recommends.

dagurval:
Allows for other methods of synchronization besides ntp. Issue #131.


dagurval:
I have created a .deb package with this change for testing at http://xt.allnodes.xyz/dev/bitcoinxt-0.11F.deb

Issue #153, opened 2016-06-30
xxx does not support thin blocks, disconnecting


KlausT:
Is this intended behavior when I don't use any special options?

> '/Satoshi:0.12.1/'` - peer=8 does not support thin blocks, disconnecting
> '/Classic:1.1.0/' - peer=11 does not support thin blocks, disconnecting

0.11.x and older clients are not being disconnected, as far as I can see

dagurval:
At around the time Bitcoin XT started doing bloom thin blocks, Core changed the standard behaviour of merkleblock requests and broke compatibility with our thin block implementation. Classic also has this one liner change. More discussion here: https://github.com/bitcoinclassic/bitcoinclassic/pull/125

For _outgoing_ connections we avoid peers that don't support thin blocks. This means core > 0.12 and derivatives. We still _accept incoming_ connections from any peers and download full blocks from them (unless configured not to).

For now, this means outgoing connections to:
- Anyone that signals support for xthin blocks (XT, BU and hopefully soon Classic)
- Older XT nodes
- Bitcoin Core <= 0.11 nodes

So in other words, this is intended behaviour.

NanoAkron:
I find any behaviour that segregates peers to be a bit distasteful. This is meant to be 'peer-to-peer electronic cash' after all.

dagurval:
Peer selection for _outgoing connections only_ is needed to be able to use the thin blocks feature of XT. Core 0.12 simply does not support thin blocks. If this is distasteful to you, please disable this feature on your node.

Anyhow, If you look at what peers XT nodes are connected to, you will see that there is none to little segregation of peers in practice. Here is my node: 

> $ ./bitcoinxt/src/bitcoin-cli getpeerinfo | grep subver | wc -l
> 94
> $ ./bitcoinxt/src/bitcoin-cli getpeerinfo | grep subver | grep "Satoshi:0.12" | wc -l
> 36
> 

38% of the nodes my node is connected to is Bitcoin Core 0.12. With a bigger sample size, that number is likely to climb closer to the Core 0.12 share of ~50% of the network nodes.

NanoAkron:
Which feature - connecting only to XThin supporting nodes, or XThin itself?

dagurval:
You cannot disable peer selection without disabling thin blocks at this time.

dgenr8:
[...]

PR #159, opened 2016-08-10
Direct headers announcement

dagurval:
Further work on #137. Incorporates #148 because I wanted to use C++11.

This work is based on @sdaftuar work in Bitcoin Core PR #6494 (9fa8c48).
This work is based on @dgenr8 cherry picks in Bitcoin XT #137
    
This implements BIP130. This implementation is a refactor/rewrite on previous work done in Bitcoin Core.
    
This work supports headersfirst together with thin blocks. It is refactored to support unittests, and unittests have been written for most of the logic.

PR #160, opened 2016-08-11
Option to enable/disable strict peer selection

dagurval:
Add option to enable/disable strict peer selection.

Added option -use-peer-selection. If this is set to 0, then we will not
require outgoing peer connections to support thin blocks.

Should solve issue https://github.com/bitcoinxt/bitcoinxt/issues/153

When there are enough thin block supporting clients on the network we should consider changing the default to 0.


dagurval:
Branched out from a commit in #159, it's probably easiest to merge that PR first.

dgenr8:
Needs rebase

dagurval:
I changed my mind on this. We don't need yet another configuration parameter. Now that Core nodes support thin blocks we connect to them as well.

I want to keep the tests though, so I added them in a new PR, #175

PR #162, opened 2016-09-01
Compact blocks (BIP152)

dagurval:
So Core went ahead and created yet another thin block protocol. This is the third thin block implementation I add to XT. It’s getting kind of repetitive. But this is a good step in the direction of on-chain scaling so I’m happy too. This PR is created on top of PR #159.

This PR has limited testing: It’s been running on my node since yesterday without issues, sharing compact blocks with Core nodes and xthin blocks with others.  I have also successfully tested all rpc-tests (except pruning.py which i skipped) with compact block transfers.

**The missing:**
- I have cherry-picked p2p-compactblocks.py by @sdaftuar, but I have not been able to get the test to run yet. There are likely issues with regtest differences because of BIP109. Help would be appreciated.
- I have not implemented _high bandwidth mode_. It is optional per BIP152. I may implement it later together with XThin Xpedited.

**The bonus:**
- Uses inventoryKnown add transactions the remote node is missing to the compact blocks sent - reducing latency. This is in contrast to the Core implementation that only sends the coinbase tx.
- Supports parallel download of the same block -- meaning no stalled downloads. XT continues to be the only full node supporting parallel download of the chain tip.
- Can parallel download xthin and bloom thins for the same block.

Compact blocks are very similar to xthin. Thus this implementation fits well within existing architecture.

### DIFF
--- a/src/blockencodings.cpp
+++ b/src/blockencodings.cpp
@@ -1,0 +1,158 @@
+// Copyright (c) 2016 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "blockencodings.h"
+#include "consensus/consensus.h"
+#include "consensus/validation.h"
+#include "chainparams.h"
+#include "hash.h"
+#include "random.h"
+#include "streams.h"
+#include "txmempool.h"
+#include "main.h"
+
+#include <unordered_map>
+
+#define MIN_TRANSACTION_SIZE (::GetSerializeSize(CTransaction(), SER_NETWORK, PROTOCOL_VERSION))
+
+CBlockHeaderAndShortTxIDs::CBlockHeaderAndShortTxIDs(const CBlock& block) :
+        nonce(GetRand(std::numeric_limits<uint64_t>::max())),
+        shorttxids(block.vtx.size() - 1), prefilledtxn(1), header(block) {
+    FillShortTxIDSelector();
+    //TODO: Use our mempool prior to block acceptance to predictively fill more than just the coinbase
+    prefilledtxn[0] = {0, block.vtx[0]};
+    for (size_t i = 1; i < block.vtx.size(); i++) {
+        const CTransaction& tx = block.vtx[i];
+        shorttxids[i - 1] = GetShortID(tx.GetHash());
+    }
+}
+
+void CBlockHeaderAndShortTxIDs::FillShortTxIDSelector() const {
+    CDataStream stream(SER_NETWORK, PROTOCOL_VERSION);
+    stream << header << nonce;
+    CSHA256 hasher;
+    hasher.Write((unsigned char*)&(*stream.begin()), stream.end() - stream.begin());
+    uint256 shorttxidhash;
+    hasher.Finalize(shorttxidhash.begin());
+    shorttxidk0 = shorttxidhash.GetUint64(0);
+    shorttxidk1 = shorttxidhash.GetUint64(1);
+}
+
+uint64_t CBlockHeaderAndShortTxIDs::GetShortID(const uint256& txhash) const {
+    static_assert(SHORTTXIDS_LENGTH == 6, "shorttxids calculation assumes 6-byte shorttxids");
+    return SipHashUint256(shorttxidk0, shorttxidk1, txhash) & 0xffffffffffffL;
+}
+
+
+
+ReadStatus PartiallyDownloadedBlock::InitData(const CBlockHeaderAndShortTxIDs& cmpctblock) {
+    if (cmpctblock.header.IsNull() || (cmpctblock.shorttxids.empty() && cmpctblock.prefilledtxn.empty()))
+        return READ_STATUS_INVALID;
+    if (cmpctblock.shorttxids.size() + cmpctblock.prefilledtxn.size() > MAX_BLOCK_SIZE / MIN_TRANSACTION_SIZE)
+        return READ_STATUS_INVALID;
+
+    assert(header.IsNull() && txn_available.empty());
+    header = cmpctblock.header;
+    txn_available.resize(cmpctblock.BlockTxCount());
+
+    int32_t lastprefilledindex = -1;
+    for (size_t i = 0; i < cmpctblock.prefilledtxn.size(); i++) {
+        if (cmpctblock.prefilledtxn[i].tx.IsNull())
+            return READ_STATUS_INVALID;
+
+        lastprefilledindex += cmpctblock.prefilledtxn[i].index + 1; //index is a uint16_t, so cant overflow here
+        if (lastprefilledindex > std::numeric_limits<uint16_t>::max())
+            return READ_STATUS_INVALID;
+        if ((uint32_t)lastprefilledindex > cmpctblock.shorttxids.size() + i) {
+            // If we are inserting a tx at an index greater than our full list of shorttxids
+            // plus the number of prefilled txn we've inserted, then we have txn for which we
+            // have neither a prefilled txn or a shorttxid!
+            return READ_STATUS_INVALID;
+        }
+        txn_available[lastprefilledindex] = std::make_shared<CTransaction>(cmpctblock.prefilledtxn[i].tx);
+    }
+
+    // Calculate map of txids -> positions and check mempool to see what we have (or dont)
+    // Because well-formed cmpctblock messages will have a (relatively) uniform distribution
+    // of short IDs, any highly-uneven distribution of elements can be safely treated as a
+    // READ_STATUS_FAILED.
+    std::unordered_map<uint64_t, uint16_t> shorttxids(cmpctblock.shorttxids.size());
+    uint16_t index_offset = 0;
+    for (size_t i = 0; i < cmpctblock.shorttxids.size(); i++) {
+        while (txn_available[i + index_offset])
+            index_offset++;
+        shorttxids[cmpctblock.shorttxids[i]] = i + index_offset;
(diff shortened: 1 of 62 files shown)


dgenr8:
dgenr8-bot reports successful build, and all tests passed.

src/blockencodings.cpp
```diff
@@ -15,21 +16,45 @@
...
+
+    //< Index of a prefilled tx is its diff from last index.
+    size_t prevIndex = 0;
+    prefilledtxn.push_back(PrefilledTransaction{0, block.vtx[0]});
+
     for (size_t i = 1; i < block.vtx.size(); i++) {
         const CTransaction& tx = block.vtx[i];
-        shorttxids[i - 1] = GetShortID(tx.GetHash());
+
+        if (inventoryKnown && !inventoryKnown->contains(tx.GetHash()))
```
gmaxwell: This at a minimum this should limit the amount it sends, and not use this heuristic when the peer has transaction relay disabled (!fRelayTxes).

BIP152 states:
"Nodes sending cmpctblock messages SHOULD limit prefilledtxn to 10KB of transactions. When in doubt, nodes SHOULD only include the coinbase transaction in prefilledtxn."

If it's going to send more than that, the latency is already hosed, and it might as well take the extra round trip and not waste bandwidth.

Filterinventoryknown isn't going to match any transactions if the peer is blocks only, so as implemented this will spam them with the whole block, and waste an epic ton of bandwidth for no good reason.

What we tested while defining the protocol was using the list of transactions which were misses on this node when it received the block. This was highly effective and had a low false positive rate, you might want to consider that. You could further augment it with leaving out things the far end has advertised-- we never tried that, but it would be an interesting experiment.

In general, I found the HB mode to be much more effective at reducing latency, with a somewhat lower bandwidth penalty than extensive prediction; but I'm glad you're experimenting here.

src/blockencodings.cpp
```diff
@@ -40,133 +65,31 @@ void CBlockHeaderAndShortTxIDs::FillShortTxIDSelector() const {
...
-    LOCK(pool->cs);
-    for (CTxMemPool::txiter it = pool->mapTx.begin(); it != pool->mapTx.end(); it++) {
-        std::unordered_map<uint64_t, uint16_t>::iterator idit = shorttxids.find(cmpctblock.GetShortID(it->GetTx().GetHash()));
-        if (idit != shorttxids.end()) {
-            if (!have_txn[idit->second]) {
-                txn_available[idit->second].reset(new CTransaction(it->GetTx()));
-                have_txn[idit->second]  = true;
-                mempool_count++;
-            } else {
-                // If we find two mempool txn that match the short id, just request it.
```
gmaxwell: Removing the collision handling cases (unless I missed where you added them elseswhere) will dramatically reduce the effectiveness of reconstruction with large mempools. The code here was driven by doing the math, and by testing it, not some lark. :) just thought you should know.

src/blockencodings.cpp
gmaxwell: I failed to mention, not just blocks only peers will get spammed, but any peer you recently connected to-- probably for the first couple hours after the connection.   Which is why I think it should just check to see if it's predicted more than a few transactions or kilobytes, and if so-- send the coinbase only.

gmaxwell:
Hi. For some odd reason my line comments seem to be vanishing moments after I make them, I wouldn't want this to get lost:

Regarding the use of filterInventoryKnown for prediction,

This at a minimum this should limit the amount it sends, and not use this heuristic when the peer has transaction relay disabled (!fRelayTxes).

BIP152 states:
"Nodes sending cmpctblock messages SHOULD limit prefilledtxn to 10KB of transactions. When in doubt, nodes SHOULD only include the coinbase transaction in prefilledtxn."

If it's going to send more than that, the latency is already hosed, and it might as well take the extra round trip and not waste bandwidth.

Filterinventoryknown isn't going to match any transactions if the peer is blocks only, so as implemented this will spam them with the whole block, and waste an epic ton of bandwidth for no good reason.  It will also not match any if the peer is newly connected-- and probably will also send significant chunks of the block for several hours.  This is why you should probably check if the predicted amount would over some size (no larger than 10kb) and if it would be don't bother predicting, just send the coinbase transaction. (at 10kb of predicted txns TCP itself is going to start resulting in multiple round trips)

What we tested while defining the protocol was using the list of transactions which were misses on this node when it received the block. This was highly effective and had a low false positive rate, you might want to consider that. You could further augment it with leaving out things the far end has advertised-- we never tried that, but it would be an interesting experiment.

In general, I found the HB mode to be much more effective at reducing latency, with a somewhat lower bandwidth penalty than extensive prediction; but I'm glad you're experimenting here.

On the reconstruction code:
Removing the collision handling cases (unless I missed where you added them elseswhere) will dramatically reduce the effectiveness of reconstruction with large mempools. The code here was driven by doing the math, and by testing it, not some lark. :) just thought you should know.

If you have any feedback on the spec, please make sure to let the author or the mailing list know.

Cheers,

src/blockencodings.cpp
dagurval: > This at a minimum this should limit the amount it sends, and not use this heuristic when the peer has transaction relay disabled (!fRelayTxes).

Good suggestion, I will add this check.

> Filterinventoryknown isn't going to match any transactions if the peer is blocks only

When does it make sense to request a compact block in blocks only mode? Anyway, this will be fixed by checking fRelayTxes (as you suggested).

> What we tested while defining the protocol was using the list of transactions which were misses on this node when it received the block. This was highly effective and had a low false positive rate, you might want to consider that.

@gandrewstone also recommended the same approach. It is what they use with Xpedited. I may change to this approach when/if I add hight bandwidth mode for both protocols.

I’ll take your advise and limit prefilledtxn to <= 10KB.

dagurval:
I don’t know why GitHub claims your comments are for an outdated diff. I have not pushed any changes since last night. I responded to you in the “outdated diff”. I hope you can find my response.

Thanks for the review @gmaxwell. It was helpful.

src/blockencodings.cpp
dagurval: I would appreciate it if you took the time to share the math.

dagurval:
Pushed https://github.com/bitcoinxt/bitcoinxt/pull/162/commits/224dd24be815eef0176d2be9cd87b0c4391d2721 for better prefilling. From the commit:

> Prefilled txs from inventory known are now limited to < 10KB.
>
> Added CompactPrefiller interface. Allows for implementation of different
> strategies for prefilling. Created a CoinbaseOnlyPrefiller strategy used
> when sending compact blocks to peers that don't want transactions
> relayd.

Does this address your concerns with regard to prefilling @gmaxwell ?

dagurval:
I added a check for hash collision in mempool. Transactions will be re-requested if there is a collision.

I also added a workaround for issue in Core https://github.com/bitcoin/bitcoin/issues/8705

I removed the [WIP] tag. My opinion is that the rpc-test port is critical for this to get merged.

dagurval:
Turns out that it was I who misunderstood the _differential encoding_. I was off by one. I've fixed the encoding/decoding and squashed it into other commits.

dagurval:
https://github.com/bitcoin/bitcoin/pull/8739

dagurval:
Tweaked p2p-compactblocks.py test and enabled it.

src/blocksender.cpp
```diff
@@ -84,21 +98,28 @@ void BlockSender::sendBlock(CNode& node,
 
     // We only support MSG_XTHINBLOCK, if peer wants MSG_THINBLOCK,
     // fallback to full one.
-    if (invType == MSG_BLOCK || invType == MSG_THINBLOCK)
+    if (invType == MSG_BLOCK || (invType == MSG_THINBLOCK
+                && !NodeStatePtr(node.id)->supportsCompactBlocks))
```
rebroad: I don't like this.... what if the node also supports compact blocks but is actually sending MSG_THINBLOCK? Let's not assume it's NOT MSG_THINBLOCK, especially if they have the XTHIN service bit set, which would suggest that it is.

dagurval:
rebased

TheBlueMatt:
Please make sure you follow the updated BIP at https://github.com/bitcoin/bips/pull/473 for releases, thanks.

dagurval:
Thanks for the heads up @TheBlueMatt 

High bandwidth mode will be done in a seperate PR, so this update does not affect this PR. But its good to know of it when I get around to adding HBM.

src/blocksender.cpp
rebroad: I mean, sending a MSG_THINBLOCK INV - it could be requesting a THINBLOCK _and_ supporting Compact Blocks.

dgenr8:
No problems on production node running 1 week. Successfully connects outbound to Core 0.13.0.

ACK

src/net.cpp
```diff
@@ -2201,3 +2201,11 @@ bool CNode::SupportsBloomThinBlocks() const {
 bool CNode::SupportsXThinBlocks() const {
     return nServices & NODE_THIN;
 }
+
+bool CNode::SupportsCompactBlocks() const {
+
+    // If a Core node activates segwit, it also
+    // disables support for compact blocks v1
+    return nVersion >= SHORT_IDS_BLOCKS_VERSION
+        && !(nServices & NODE_WITNESS);
```
dagurval: You're right. I was under the impression that the flag would be set after activation. However, it is  set if activation criteria is defined: https://github.com/bitcoin/bitcoin/blob/ed8d693c71b01caa310922f883b3915c09bb6a35/src/init.cpp#L1487

Also, that comment/criteria is based on old Core behavior. I believe they have added support for v1 after segwit activation. We only need to check nVersion now. Good find @rebroad , thanks

PR #148, opened 2016-05-05
C++11

dagurval:
Cherry-picks from https://github.com/bitcoinclassic/bitcoinclassic/pull/162

PR #163, opened 2016-09-12
Split repo - bump to C+11

dagurval:
This is an alternative to #148 

This PR syncs doc, contrib and depends with Core. It also bumps XT to C++11. I have successfully built gitian builds for GNU/Linux and Windows with this PR.

I have split off "doc", "contrib" and "depends" into separate repos. Non of these directories are required to build Bitcoin XT. They are only required for Gitian.

I have merged changes from Core into these repos. They are now up-to-date with Core. The reasoning behind splitting the repos like this is that it will be easier to maintain these folders separate, rather than trying to cherry pick them up-to-date in the main repo.

I also made a script to check out/pull updates of these folders:
https://github.com/dagurval/bitcoinxt/blob/split-repo/checkout-all.pl

The sub-repos are at:
https://github.com/dagurval/bitcoinxt-doc
https://github.com/dagurval/bitcoinxt-contrib
https://github.com/dagurval/bitcoinxt-depends

These should ideally be cloned into the bitcoinxt project.

I've cherry-picked relevant commits in build-aux. This enabled c++11 build.
https://github.com/dagurval/bitcoinxt/commits/split-repo/build-aux/

I cherry picked some C++11 changes from PR #148 :
* replace auto_ptr with unique_ptr  
* use c++11 syntax instead of boost::list_of
* Use boost reverse lock instead of own
And added a workaround for boost c++11 issue.

XT specific changes for Gitian build:
=====================================
* Revert removal of a java comparison tool that we still use:
https://github.com/dagurval/bitcoinxt-depends/commit/beca65dc008a3c014f0967d20900fcb2729290ab

* Check-out step for repos split out in gitian:
https://github.com/dagurval/bitcoinxt-contrib/commit/f8970fdf0542c2f7bc66d2a551237ab81679b387

* Disabled symbol checking
XT exports some symbols that Core does not. I have disabled it for now. This check is to automatically check that the binaries are compatible with old Debian and Ubuntu releases. For now we need to manually check if Bitcoin XT can run on latest Debian and Ubuntu (or older, if we care to support older)
https://github.com/dagurval/bitcoinxt-contrib/commit/b0b647e48cdbc4486f91eec7f59c504bf958fd8a

* Disabled all arch except for 32bit and 64bits in Linux. I had linker issues with arm.
https://github.com/dagurval/bitcoinxt-contrib/commit/cac8d1aa0ba39107bd3c6a7cb772e4373a6edf2a


dagurval:
Closed in favour of #164

PR #167, opened 2016-10-30
Qt: Add option to hide the system tray icon

rimmington:
This is a cherry-pick from Bitcoin Core. I'm requesting it be merged in here because a) I'm getting odd tray icon behaviour with the official builds on Ubuntu/Unity and b) I've got enough tray icons already.

Merged cleanly and works in my informal tests on Ubuntu.


dagurval:
Thanks.

I compiled and ran on OS X, but turns out tray is always disabled on OS X 😅.

So, utACK

dgenr8:
ACK

PR #178, opened 2016-12-07
Make net debugging more granular

rebroad:
debug=net tends to produce too much output and it can often be desired to debug just certain parts of what is provided by net debug. I have split it down into the following:-

This wll help with when developing just certain areas, e.g. block propagation, and therefore we're not needing visibility of tx send/get logic.

As yet the command line help is not updated - can add this if it is wanted (or a maintainer can).

"block" debugs block download progress
"blocksend" debugs block upload progress
"netrecv" debugs every message received
"netsend" debugs every message sent
"conn" debugs the making and breaking of connections to other peers
"neterr" debugs network errors
"net" everything else

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -372,7 +372,7 @@ CNode* ConnectNode(CAddress addrConnect, const char *pszDest)
     }
 
     /// debug print
-    LogPrint("net", "trying connection %s lastseen=%.1fhrs\n",
+    LogPrint("conn", "trying connection %s lastseen=%.1fhrs\n",
         pszDest ? pszDest : addrConnect.ToString(),
         pszDest ? 0.0 : (double)(GetAdjustedTime() - addrConnect.nTime)/3600.0);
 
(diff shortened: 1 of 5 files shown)


src/net.cpp
```diff
@@ -2112,7 +2112,7 @@ void CNode::AskFor(const CInv& inv)
         nRequestTime = it->second;
     else
         nRequestTime = 0;
-    LogPrint("net", "askfor %s  %d (%s) peer=%d\n", inv.ToString(), nRequestTime, DateTimeStrFormat("%H:%M:%S", nRequestTime/1000000), id);
+    LogPrint("tx", "askfor %s  %d (%s) peer=%d\n", inv.ToString(), nRequestTime, DateTimeStrFormat("%H:%M:%S", nRequestTime/1000000), id);
```
dgenr8: Is this only ever reached for inv items that are txes?

dgenr8:
This looks like a good change to me.  Any help that is obsoleted should be part of it.

dagurval:
I also like this. However, this will change the log output in the rpc-tests.

When an rpc-test fails, it is difficult to know what to look for. Having all output for “net” is helpful.

Is it possible to for example add a “netall” that enables output for all these? rpc-tests that enable “net” today should then be updated to enable “netall” as well.

dagurval:
I checked the rpc-tests. Only two tests use -debug=net, the others use -debug which enables all. So never mind my previous comment.

dgenr8:
Closing until help and "netall" can be addressed.

PR #181, opened 2016-12-22
Revert BIP109

dgenr8:
This leaves the repository in a state without any hard fork support.


src/miner.cpp
```diff
@@ -138,27 +147,6 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
                                 ? nMedianTimePast
                                 : pblock->GetBlockTime();
 
-        UpdateTime(pblock, Params().GetConsensus(), pindexPrev);
```
dagurval: Is UpdateTime part of BIP109?

dgenr8: This call to UpdateTime was added in BIP101, and was kept when BIP101 was reverted and BIP109 merged.

dagurval:
ACK

PR #182, opened 2016-12-29
Rewrite CreateNewBlock

dgenr8:
Based on https://github.com/bitcoin/bitcoin/pull/6898 by @morcos which appeared in core 0.12.0.

Use the fee score index on the mempool to add transactions to a new block in sorted order.  XT's mempool policy, which strongly discourages unconfirmed chains by expecting a child to pay for all unconfirmed ancestors, is extended to CreateNewBlock by mining these paid-for ancestors.

Removes much of the validation while building the block, relying on mempool to be consistent and only contain txs that can be mined.

The old "priority" block area is removed and minimum block size is no longer supported.

Individual transactions may be excluded by setting a fee delta that drops them below ::MinRelayFee, as expected by tests.  The -relaypriority=0 argument is supported as a way to cause free or effectively-free transactions to be mined.  

As a reminder, XT's minimum relay fee is fixed, currently 1000 satoshis.  There is no dynamic minimum or exponential decay thereof, and mempool ejection is done on 72-hour expiration or when full, by lowest fee score as described above.

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -49,10 +49,9 @@ class CValidationState;
 struct CNodeStateStats;
 struct LockPoints;
 
-/** Default for -blockmaxsize and -blockminsize, which control the range of sizes the mining code will create **/
+/** Default for -blockmaxsize, which controls the range of sizes the mining code will create **/
 static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 750000;
-static const unsigned int DEFAULT_BLOCK_MIN_SIZE = 0;
-/** Default for -blockprioritysize, maximum space for zero/low-fee transactions **/
+/** Used only for relay rule that guesses at network miner policies **/
 static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = 50000;
 /** The maximum size for transactions we're willing to relay/mine */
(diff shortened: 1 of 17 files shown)


dgenr8:
Bugfix pushed.

Incidentally, testing on mainnet shows that getblocktemplate is 4X-10X as fast as the old 11.0-based version, on live mempools.  I'm not sure how it compares to current core versions with their relatively complex package management.

dgenr8:
Bugfix pushed. TestBlockValidity could fail if an unconfirmed parent was non-final.
@haipome @jtoomim

src/main.h
```diff
@@ -255,6 +254,8 @@ void PruneAndFlush();
 bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransaction &tx, bool fLimitFree,
                         bool* pfMissingInputs, bool fOverrideMempoolLimit=false, bool fRejectAbsurdFee=false);
 
+/** Convert CValidationState to a human-readable message for logging */
+std::string FormatStateMessage(const CValidationState &state);
```
dgenr8: Not worth it because of future rebasing brain cycles.

PR #183, opened 2016-12-29
Reduce default number of blocks to check at startup

dgenr8:
Cherry-pick of https://github.com/bitcoin/bitcoin/pull/8611.

This speeds startup by reducing the default number of recent blocks to re-check.


dagurval:
utACK

PR #188, opened 2017-03-13
BIP100: Dynamic maximum block size by miner vote

dgenr8:
These changes are for Bitcoin XT. For the Bitcoin Core version, see bitcoinxt/bitcoin#1.

Implementation of the [BIP100 specification](https://github.com/jgarzik/bip100/blob/master/bip-0100.mediawiki) as published.

Full nodes need no configuration to follow the network sizelimit as determined by BIP100.  Miners may set their coinbase /B vote using 

``-maxblocksizevote=<n> Set vote for maximum block size in megabytes (default: network sizelimit)``

Most RPC's return the sizelimit as of the current tip, consistent with the other attributes.  ``getblocktemplate`` returns the sizelimit for a block built on the current tip, which may change by a factor of up to 1.05 at the start of a difficulty interval, per the specification.

On first run, block index entries starting at the activation height (449568, which was in January 2017) are updated to track miner size votes and sizelimit history.  Use ``-debug=reindex`` (NOT ``-reindex``) to see the progress of this update.

The current network sizelimit is published in the user agent string as EB.

This implementation does not lift the 32MB physical MAX_SIZE limit.  File buffer sizes, sanity checks in bitcoin-tx, verifytxoutproof, and merkle blocks also continue to reference MAX_BLOCK_SIZE as a general scale indicator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,7 +105,8 @@ linux-build
 win32-build
 qa/pull-tester/run-bitcoind-for-test.sh
 qa/pull-tester/tests-config.sh
-qa/pull-tester/cache/*
+qa/pull-tester/cache
+qa/pull-tester/cache_bigblock
 qa/pull-tester/test.*/*
 
 !src/leveldb*/Makefile
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -96,6 +96,7 @@ testScripts = [
     'sendheaders.py'
 ]
 testScriptsExt = [
+    'bip100-sizelimit.py',
     'bip9-softforks.py',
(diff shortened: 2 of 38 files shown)


dgenr8:
Pushed a parser bugfix from @dagurval.  We'll squash these after a while.

dgenr8:
Pushed changes to support specification update https://github.com/jgarzik/bip100/pull/20.

dgenr8:
Pushed coinbaseaux support from @dagurval.

dgenr8:
Pushed vote search edge cases and some squashes.

dgenr8:
Added compact blocks update from @dagurval.

dagurval:
ACK

PR #192, opened 2017-04-06
Generic method to request preceding headers.

dagurval:
Always request preceding headers for headers, blocks and thinblocks received. If preceding headers were missing, discard the received block (to be downloaded again later, if we still want it). 

This is mostly done already, but in a more ad-hoc fasion, so this simplifies the code.


dgenr8:
Rebased.

dgenr8:
ACK

PR #194, opened 2017-04-07
Announce blocks with compact blocks

dagurval:
Adds support for announcing blocks using compact blocks instead of inv or header. Remote node must requests "high bandwith mode" compact blocks to enable this.

Note: XT does not request block announcements yet, but can now serve them.


dgenr8:
utACK

PR #198, opened 2017-04-25
Remove option -logips. Re-add logging of IPs.

dagurval:
We should either remove or or maintain this feature. I prefer for less code (remove).

`-logips` was adedded in https://github.com/bitcoin/bitcoin/pull/3764


dagurval:
If we want to maintain instead:
* ID needs to be added to "peer misbehaved" output. Currently shows only IP. Can't know what peer was misbehaved without adding logips=1
* IPGroup name needs to be obfuscated, default name for IPGroup is the IP address. IPGroup is logged when a peer connects.

dgenr8:
Rebased.

dgenr8:
I think we should maintain this option.  It's nice to be able to share log snippets without worrying about revealing your peer IP's.

dagurval:
Closing for now

PR #200, opened 2017-05-10
Update DNS seeds

dgenr8:


dgenr8:
Seed removed. Thanks @dagurval.  I'll advise BU as well.

jgarzik:
posthumous ACK

PR #204, opened 2017-05-20
Don't keep outgoing connections to UASF nodes.

dagurval:
UASF nodes reject blocks that are valid, but do not vote for a consensus change the node operator is trying to force onto the network. Avoid UASF nodes to always see the best chain.

These nodes can still have incoming connections to us.


jonasschnelli:
ACK

PR #207, opened 2017-06-01
Remove bloom thin blocks

dagurval:
XT was first to implement thin blocks and did so using backwards compatible protocol calls - merkleblocks.

This protocol has its downsides and has been superseeded by now widely deployed xthin and compact blocks.

It's time to remove bloom thin blocks to reduce code maintenance.


dagurval:
The big diff in ProcessMessage->tx is just indentation change after removing the `tryMempool` variable.

dagurval:
Pushed a commit to fallback to full block download when re-requesting compact/xthin transactions does not provide all the transcations we're missing. Regression was caught by extended rpc tests.

dgenr8:
ACK

PR #203, opened 2017-05-18
Networking improvements

dgenr8:
Networking, peer management, and privacy cherry-picks.

addrman catch-up
https://github.com/bitcoin/bitcoin/pull/7125 Replace global trickle node with random delays
https://github.com/bitcoin/bitcoin/pull/6412 Test whether created sockets are select()able
  https://github.com/bitcoin/bitcoin/pull/7696 Fix de-serialization bug where AddrMan is left corrupted
https://github.com/bitcoin/bitcoin/pull/8065 Addrman offline attempts
https://github.com/bitcoin/bitcoin/pull/8282 Feeler connections to increase online addrs in the tried table
https://github.com/bitcoin/bitcoin/pull/8944 Remove bogus assert on number of oubound connections
https://github.com/bitcoin/bitcoin/commit/d74e352e0136bb746c2975f9a4092f23b9d8899c Set feelers to disconnect at the end of the version message
https://github.com/bitcoin/bitcoin/pull/8069 Remove non-determinism which is breaking net_tests
https://github.com/bitcoin/bitcoin/pull/7932 CAddrMan::Deserialize handle corrupt serializations better

PR #208, opened 2017-06-08
Thin block announcements

dagurval:
This adds support for thin block announcements and implements BIP152's _high bandwith mode_. The rpc test changes are ports from Core with the segwit parts of it disabled.

### DIFF
--- a/qa/rpc-tests/p2p-compactblocks.py
+++ b/qa/rpc-tests/p2p-compactblocks.py
@@ -648,6 +648,8 @@ class CompactBlocksTest(BitcoinTestFramework):
         node.generate(1)
         wait_until(test_node.received_block_announcement, timeout=30)
         test_node.clear_block_announcement()
+        with mininode_lock:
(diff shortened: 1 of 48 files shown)


dgenr8:
``bipdersig-p2p.py``, ``bip65-cltv-p2p.py`` result in "Assertion failed: Not all nodes requested block"

Rebases cleanly onto #203, we must have pleased the gods somehow.

dagurval:
Behaviour was changed to download and validate headers before downloading full block on inv announcements. These tests assumed that full block would always be downloaded.

Reverted back to always download full blocks  https://github.com/bitcoinxt/bitcoinxt/pull/208/commits/f42cc265d7a208841618fedfd9e121c7cb1ddde1. Had to tweak `sendheaders` test due to differences in batching getdata calls. This commit can be squashed.

All tests pass for me now.

dgenr8:
ACK with minimal manual testing.

PR #209, opened 2017-06-28
Fix qt/paymentrequestplus.cpp for OpenSSL 1.1 API

dgenr8:
https://github.com/bitcoin/bitcoin/pull/9326 Fix qt/paymentrequestplus.cpp for OpenSSL 1.1 API


dagurval:
utACK

PR #210, opened 2017-06-29
Fix timing issue in txn_clone.py

dgenr8:
Fix timing issue in txn_clone.py by whitelisting peers.


dgenr8:
Closed in favor of #213

PR #212, opened 2017-07-03
util: Update tinyformat

dagurval:

PR #214, opened 2017-07-06
Add missing arguments to log statement

dagurval:
Fixes test regression in https://github.com/bitcoinxt/bitcoinxt/pull/212


dgenr8:
That does the trick.

PR #216, opened 2017-07-11
Increase timeout in txn_doublespendrelay

dagurval:
The test txn_doublespendrelay is randomly failing.

After investigating I can see that the double spend reporting works, however transactions some times take a very long time to be passed between nodes.

Increasing the timeout makes the test pass. I have yet to find out why transactions are slow to be passed between nodes.

This PR also makes the rpc test state explicitly when it fails due to timeout.


dgenr8:
ACK.  Awesome @dagurval!

PR #219, opened 2017-07-20
UAHF: REQ 6-2, REQ 6-3. Other-chain replay protection.

dgenr8:
After the fork, allow transactions with SIGHASH_FORKID set, protecting them from replay on other chains.

Also implement signature hash algorithm improvements per the UAHF [sub-specification](https://github.com/Bitcoin-UAHF/spec/blob/master/replay-protected-sighash.md).


dgenr8:
Closed in favor of #226.

PR #223, opened 2017-07-24
UAHF: REQ-7 Difficulty adjustement in case of hashrate drop

dgenr8:
Between regular adjustments, difficulty drops 20% when timestamp difference >= 12 hours is seen across 6 block intervals.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -439,6 +439,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-maxblocksizevote=<n>", _("Set vote for maximum block size in megabytes (default: network sizelimit)"));
     if (showDebug)
         strUsage += HelpMessageOpt("-blockversion=<n>", "Override block version to test forking scenarios");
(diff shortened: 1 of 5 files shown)


dgenr8:
This change accepts the risk of an attack that I analyzed as follows.

Miners can crank the difficulty down by following this strategy:

    Slow-walk MTP in 1-second increments
    When MTP is 10 hours behind clock time, mine a block with timestamp = clock time + 2 hours

The final block will have MTP about 12 hours greater than the previous block (and therefore the one 5 before that).

About 72% of hashpower is sufficient for this strategy to have a 50% chance of working each time it is tried, but that it would be very unlikely to work with only 40% hashpower.  Only the percentage matters -- it would work with any absolute hashrate.

The above is a worst-case scenario for the attacker, in the sense it assumes maintaining the timewarp for 10 hours requires 60 blocks with no MTP reset. If we assume a sudden drop to 33.3% of pre-fork rate at UAHF activation, a miner with 40% of that has nearly a 5% chance of pulling it off each time he tries.  At about 10% of pre-fork rate, the change is designed to trigger anyway.

These are analytical results. If they are of any interest, it would be good to confirm them by simulation.

Issue #235, opened 2017-08-02
0.11.0G Node cannot join Bitcoin Cash Fork


chaosgrid:
My node gets stuck on block 478558 and flags any nodes that seemingly want to send me the rightful block as Misbehaving:

`2017-08-01 18:13:08 received compactblock 000000000000000000651ef99cb9fcbe0dadde1d424bd9f15ff20136191a5eec from peer=229
2017-08-01 18:13:08 received compactblock 000000000000000000651ef99cb9fcbe0dadde1d424bd9f15ff20136191a5eec from peer=681
2017-08-01 18:13:09 Misbehaving: 5.39.77.33:8333 (0 -> 20)
2017-08-01 18:13:09 Misbehaving: 39.108.71.254:8333 (0 -> 20)
2017-08-01 18:13:10 received compactblock 000000000000000000651ef99cb9fcbe0dadde1d424bd9f15ff20136191a5eec from peer=233
2017-08-01 18:13:11 Misbehaving: 213.105.73.173:8333 (0 -> 20)
2017-08-01 18:13:11 receive version message: /Bitcoin ABC:0.14.6(EB8.0)/: version 70015, blocks=478558, us=[2002:4e2e:6d7c::4e2e:6d7c]:50631, peerid=2716, ipgroup=2405:6580:2220:3300:822:fcac:64cb:cfb8
2017-08-01 18:13:11 receive version message: /bitnodes.21.co:0.1/: version 70015, blocks=478558, us=54.91.170.49:59293, peerid=2717, ipgroup=54.91.170.49
2017-08-01 18:13:12 receive version message: /bitnode.lee/: version 70015, blocks=478601, us=34.209.237.5:3009, peerid=2718, ipgroup=34.209.237.5
2017-08-01 18:13:13 receive version message: /bitnodes.lee.co:0.1/: version 70015, blocks=478601, us=34.211.145.37:53634, peerid=2719, ipgroup=34.211.145.37
2017-08-01 18:13:14 received compactblock 000000000000000000651ef99cb9fcbe0dadde1d424bd9f15ff20136191a5eec from peer=235
2017-08-01 18:13:14 Misbehaving: 98.223.11.58:8333 (0 -> 20)`

After restarting the node, it still gets stuck on the same block and I still get Misbehaving messages (it seems to fail on the block header already?)

[...]

PR #236, opened 2017-08-02
UAHF: Adjust block size DoS limit

dagurval:
DoS limits for block size was adjusted for BIP100, but not a sudden jump to 8MB.


dagurval:
This solves issue reported here https://github.com/bitcoinxt/bitcoinxt/issues/235

dgenr8:
ACK

PR #238, opened 2017-08-09
Add comptool.RejectResult

dagurval:
I cherry-picked this to get the class comptool.RejectResult, which is a prerequisite for a rpc test I'm porting for Bitcoin ABC.

As a bonus this also contains a couple of bug fixes and a new rpc test, invalidtxrequest.py.

Contains commits from:
https://github.com/bitcoin/bitcoin/pull/6465 - Don't share objects between TestInstances
https://github.com/bitcoin/bitcoin/pull/6509 - Fix race condition on test node shutdown
https://github.com/bitcoin/bitcoin/pull/7179 - net: Fix sent reject messages for blocks and transactions

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -93,7 +93,9 @@ testScripts = [
     'walletbackup.py',
     'p2p-versionbits-warning.py',
     'decodescript.py',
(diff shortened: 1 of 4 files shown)


dgenr8:
Awesome.

PR #239, opened 2017-08-10
Fix mempool_limit test - sign with FORKID

dgenr8:
This XT-specific test was missed in #237.


dagurval:
ACK

dgenr8:
Also fixed pruning.py.

PR #240, opened 2017-08-10
UAHF: Move check for fork activation to a function 

dagurval:
Regtests fork at genesis, so the check in ConnectTip needs to support this.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -172,6 +172,7 @@ BITCOIN_CORE_H = \
   utilstrencodings.h \
   utiltime.h \
   utilprocessmsg.h \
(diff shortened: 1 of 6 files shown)


dgenr8:
Could we also use the new function [here](https://github.com/dagurval/bitcoinxt/blob/a3b7eca93a2941b293e46cd9180ad88ce83b4849/src/main.cpp#L3447)?

dagurval:
Added unittests for the new function.

dgenr8:
Looks good. All tests pass. I'm syncing to a fork point on testnet now.

dgenr8:
ACK

PR #243, opened 2017-08-15
Add p2p-fullblocktest.py

dagurval:
Contains:
https://github.com/bitcoin/bitcoin/pull/6523 - Add p2p-fullblocktest.py
https://github.com/bitcoin/bitcoin/pull/7226 - Tests: Add more tests to p2p-fullblocktest


dgenr8:
Wow, awesome!  bipdersig-p2p.py needs to be added back to rpc-tests.py.

dagurval:
`bipdersig-p2p.py` fails because DERSIG rules are enforced in the 75th version 3 block post-fork. If you comment out `unDERify(spendtx)` it will pass. Fixing that test can be done in another PR.

dagurval:
I can re-produce the` bip9-softforks` error on both this PR and current master.

Looks like the test checks that blocks are requested + a ping-pong sync. However, occasionally UpdateTip has not run yet on the blocks received. So there is a race issue in comptools.

dgenr8:
Ok, my only request would be to remove the change to ``bipdersig-p2p.py`` since they can't be tested here.

dagurval:
Pushed a commit that fixes `bipdersig-p2p` and re-enables it.

dgenr8:
ACK

PR #248, opened 2017-08-20
Enable bitcoincash: URI prefix

dgenr8:
Cherry-pick from https://github.com/bitcoinclassic/bitcoinclassic/commit/d66afe37e762d47ea581315c6d766274d03a7b73

PR #252, opened 2017-08-24
[qt] initialize chainparams in uri test

dagurval:
UAHF status is checked to determine if prefix is `bitcoin:` or `bitcoincash:`. This requires chainparams to be initialized.

Regression was in https://github.com/bitcoinxt/bitcoinxt/pull/248


dgenr8:
ACK

PR #255, opened 2017-08-25
[debian] Add configuration step to select network

dagurval:
User is asked to select which network (BCC or BTC) to connect to when
installing the debian package.

screenshot from 2017-08-26 00-18-03

### DIFF
--- a/contrib/gitian-debian/postinst
+++ b/contrib/gitian-debian/postinst
@@ -36,3 +36,14 @@ if [ -x "/bin/systemctl" ]; then
 else
     echo "The file '/bin/systemctl' is not executable or found, bitcoinxtd service not automatically enabled or started" 
 fi
+
+# ask which network to connect to
+db_input high bitcoinxt/select_network || true
+db_go || true
+db_get bitcoinxt/select_network
+if [ "$RET" = "Bitcoin (BTC)" ]; then
+    echo "uahftime=0" >> /etc/bitcoinxt/bitcoin.conf
+    echo "BTC Network: uahftime=0 added to /etc/bitcoinxt/bitcoin.conf"
+else
+    echo "BCC Network: Default settings kept."
+fi
--- a/contrib/gitian-debian/postinst
+++ b/contrib/gitian-debian/postinst
@@ -3,8 +3,13 @@
 # Source debconf library.
 . /usr/share/debconf/confmodule
 
-# add random rpc password to bitcoin.conf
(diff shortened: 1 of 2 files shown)


dagurval:
@stevemyers You've done work on the debian package before. Could you review this pull request? 😇

dagurval:
There was a bug where every time the installer ran, we'd append a new `rpcpassword` and `uahftime` setting. Pushed a commit to fix that.

dgenr8:
Should something like ``printf "\nuahftime=0\n"`` be done in case EOF lacks a newline?

contrib/gitian-debian/postinst
```diff
@@ -42,8 +47,18 @@ db_input high bitcoinxt/select_network || true
 db_go || true
 db_get bitcoinxt/select_network
 if [ "$RET" = "Bitcoin (BTC)" ]; then
-    echo "uahftime=0" >> /etc/bitcoinxt/bitcoin.conf
-    echo "BTC Network: uahftime=0 added to /etc/bitcoinxt/bitcoin.conf"
+    if grep -q '^uahftime' $BITCOIN_CONF; then
+        sed -i 's/^uahftime.*/uahftime=0/' $BITCOIN_CONF
```
dgenr8: Why not match ^uahftime=.* similar to the matching on rpcpassword?

contrib/gitian-debian/postinst
```diff
@@ -42,8 +47,18 @@ db_input high bitcoinxt/select_network || true
...
+        echo "BTC selected: Setting uahftime=0 in $BITCOIN_CONF"
+    else
+        echo "uahftime=0" >> $BITCOIN_CONF
+        echo "BTC selected: Adding uahftime=0 to $BITCOIN_CONF"
+    fi;
 else
-    echo "BCC Network: Default settings kept."
+    if grep -q '^uahftime' $BITCOIN_CONF; then
+        sed -i 's/^uahftime.*//' $BITCOIN_CONF
```
dgenr8: ditto

contrib/gitian-debian/postinst
dagurval: For consistency `grep -q '^uahftime'` could be `grep -q '^uahftime='`. I'd have to re-test all the execution paths again, so I don't think it's worth changing.

dagurval:
> Should something like `printf "\nuahftime=0\n"` be done in case EOF lacks a newline?

`echo "uahftime=0" >> $BITCOIN_CONF` already prepends a newline and a newline is not required at the EOF

dgenr8:
utACK

Issue #256, opened 2017-08-29
pruning test fails


dagurval:
pruning fails with following error:

```
Running 2nd level testscript pruning.py ...
Initializing test directory /tmp/testu7vxbt8k
Warning! This test requires 4GB of disk space and takes over 30 mins (up to 2 hours)
Mining a big blockchain of 995 blocks
Check that we haven't started pruning yet because we're below PruneAfterHeight
Success
Though we're already using more than 550MiB, current usage: 593.0
Mining 25 more blocks should cause the first block file to be pruned
Assertion failed: blk00000.dat not pruned when it should be
```

I added the following debug output to see what's happening:

```
diff --git a/src/main.cpp b/src/main.cpp
index 8f7b115..cfc798d 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3735,19 +3735,29 @@ void FindFilesToPrune(std::set<int>& setFilesToPrune)
     int count=0;
 
     if (nCurrentUsage + nBuffer >= nPruneTarget) {
+        LogPrint("prune", "prune: Last block file: %d\n", nLastBlockFile);
         for (int fileNumber = 0; fileNumber < nLastBlockFile; fileNumber++) {
             nBytesToPrune = vinfoBlockFile[fileNumber].nSize + vinfoBlockFile[fileNumber].nUndoSize;
 
-            if (vinfoBlockFile[fileNumber].nSize == 0)
+            if (vinfoBlockFile[fileNumber].nSize == 0) {
+                LogPrint("prune", "prune: file %d already pruned\n", fileNumber);
                 continue;
+            }
 
             if (nCurrentUsage + nBuffer < nPruneTarget)  // are we below our target?
+            {
+                LogPrint("prune", "prune: Below target\n");
                 break;
+            }
 
             // don't prune files that could have a block within MIN_BLOCKS_TO_KEEP of the main chain's tip but keep scanning
             if (vinfoBlockFile[fileNumber].nHeightLast > nLastBlockWeCanPrune)
+            {
+                LogPrint("prune", "prune: Can't prune %d, height last: %d, can prune <= %d\n",
+                        fileNumber, vinfoBlockFile[fileNumber].nHeightLast, nLastBlockWeCanPrune);
                 continue;
-
+            }
+            LogPrint("prune", "prune: Marking %d for pruning\n", fileNumber);
             PruneOneBlockFile(fileNumber);
             // Queue up the files for removal
[...]

PR #265, opened 2017-09-26
Fix pruning.py failure with default 8MB

dgenr8:
Regtest is changed to have the old blockfile size of 128MB instead of the post-fork size of 1024MB, so the tuned string of tests in pruning.py continues to work.

This fixes issue #256.

### DIFF
--- a/src/chainparams.h
+++ b/src/chainparams.h
@@ -64,6 +64,8 @@ public:
     /** Policy: Filter transactions that do not match well-defined patterns */
     bool RequireStandard() const { return fRequireStandard; }
     int64_t PruneAfterHeight() const { return nPruneAfterHeight; }
+    /** Minimum number of max-sized blocks in blk?????.dat files */
+    int64_t MinBlockFileBlocks() const { return nMinBlockfileBlocks; }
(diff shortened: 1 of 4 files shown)


src/main.h
```diff
@@ -71,8 +71,6 @@ static const unsigned int DEFAULT_DESCENDANT_SIZE_LIMIT = 101;
 static const unsigned int DEFAULT_MAX_MEMPOOL_SIZE = 300;
 /** Default for -mempoolexpiry, expiration time for mempool transactions in hours */
 static const unsigned int DEFAULT_MEMPOOL_EXPIRY = 72;
-/** Minimum number of max-sized blocks in blk?????.dat files */
```
dagurval: We may want to keep the comment

dagurval:
Looks good! utACK

src/chainparams.h
```diff
@@ -64,6 +64,8 @@ class CChainParams
     /** Policy: Filter transactions that do not match well-defined patterns */
     bool RequireStandard() const { return fRequireStandard; }
     int64_t PruneAfterHeight() const { return nPruneAfterHeight; }
+    /** Minimum number of max-sized blocks in blk?????.dat files */
```
dgenr8: Comment added

PR #266, opened 2017-10-03
[Test] Update bip100-sizelimit.py for uahf

dgenr8:
Update bip100-sizelimit.py to run in the post-uahf (Bitcoin Cash fork) regtest environment.  Include cherry-picks that factor out standard big block and transaction routines, and use those routines.

### DIFF
--- a/qa/rpc-tests/bip100-sizelimit.py
+++ b/qa/rpc-tests/bip100-sizelimit.py
@@ -8,12 +8,19 @@
 #
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
+from test_framework.mininode import MAX_BLOCK_SIZE
 
 from decimal import Decimal
-
 CACHE_DIR = "cache_bigblock"
 
 class BigBlockTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 4
+        self.setup_clean_chain = None
+        # Use predictable timestamps on subsequent runs from cache
+        enable_mocktime()
 
     def setup_chain(self):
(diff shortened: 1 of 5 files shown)


qa/rpc-tests/mempool_limit.py
```diff
@@ -104,6 +51,7 @@ def run_test(self):
         txFS = self.nodes[0].signrawtransaction(tx, None, None, "NONE|FORKID")
         txid0 = self.nodes[0].sendrawtransaction(txFS['hex'])
         self.nodes[0].lockunspent(True, [us0])
+        time.sleep(3)
```
dagurval: why?

qa/rpc-tests/bip100-sizelimit.py
```diff
@@ -10,11 +10,18 @@
 from test_framework.util import *
 
 from decimal import Decimal
-
 CACHE_DIR = "cache_bigblock"
+UAHF_INITIAL_MAX_BLOCK_SIZE = 8000000
```
dagurval: Could be replaced with `from test_framework.mininode import MAX_BLOCK_SIZE`

dagurval:
Awesome. ACK!

qa/rpc-tests/bip100-sizelimit.py
dgenr8: Done

qa/rpc-tests/mempool_limit.py
dgenr8: Good catch, this was not necessary.

However, in more thorough testing I found the existing sleep had to be increased to 9 to ensure the test would succeed 30 times in a row.

PR #273, opened 2017-10-27
Encapsulate CLevelDB iterators

dagurval:
Commit from https://github.com/bitcoin/bitcoin/pull/6777, but with db obfuscation removed.

### DIFF
--- a/src/leveldbwrapper.cpp
+++ b/src/leveldbwrapper.cpp
@@ -92,3 +92,8 @@ bool CLevelDBWrapper::WriteBatch(CLevelDBBatch& batch, bool fSync) throw(leveldb
 
 // Taken from future release of DBWrapper
 const std::string CLevelDBWrapper::OBFUSCATE_KEY_KEY("\000obfuscate_key", 14);
+
+CLevelDBIterator::~CLevelDBIterator() { delete piter; }
+bool CLevelDBIterator::Valid() { return piter->Valid(); }
+void CLevelDBIterator::SeekToFirst() { piter->SeekToFirst(); }
+void CLevelDBIterator::Next() { piter->Next(); }
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -105,7 +105,7 @@ bool CCoinsViewDB::GetStats(CCoinsStats &stats) const {
        only need read operations on it, use a const-cast to get around
        that restriction.  */
     std::unique_ptr<CLevelDBIterator> pcursor(const_cast<CLevelDBWrapper*>(&db)->NewIterator());
-    pcursor->Seek('c');
+    pcursor->Seek(DB_COINS);
 
     CHashWriter ss(SER_GETHASH, PROTOCOL_VERSION);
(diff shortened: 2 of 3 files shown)


dgenr8:
The discussion mentions this a cleanup of abstractions that were missed in an earlier change.  Do we already have those as part of the obfuscation backport we included?

dagurval:
I skipped anything related to obfuscation. We still detect when a database is obfuscated, but do not support it.

src/leveldbwrapper.cpp
```diff
@@ -92,3 +92,8 @@ bool CLevelDBWrapper::WriteBatch(CLevelDBBatch& batch, bool fSync) throw(leveldb
 
 // Taken from future release of DBWrapper
 const std::string CLevelDBWrapper::OBFUSCATE_KEY_KEY("\000obfuscate_key", 14);
+
+CLevelDBIterator::~CLevelDBIterator() { delete piter; }
+bool CLevelDBIterator::Valid() { return piter->Valid(); }
+void CLevelDBIterator::SeekToFirst() { piter->SeekToFirst(); }
+void CLevelDBIterator::Next() { piter->Next(); }
```
dgenr8: Why no SeekToLast() and Prev() as in original commit?

src/txdb.cpp
```diff
@@ -104,31 +104,19 @@ bool CCoinsViewDB::GetStats(CCoinsStats &stats) const {
     /* It seems that there are no "const iterators" for LevelDB.  Since we
        only need read operations on it, use a const-cast to get around
        that restriction.  */
-    boost::scoped_ptr<leveldb::Iterator> pcursor(const_cast<CLevelDBWrapper*>(&db)->NewIterator());
-    pcursor->SeekToFirst();
+    std::unique_ptr<CLevelDBIterator> pcursor(const_cast<CLevelDBWrapper*>(&db)->NewIterator());
```
dgenr8: Was the change to unique_ptr to remove a warning?

src/txdb.cpp
```diff
@@ -190,26 +180,17 @@ bool CBlockTreeDB::ReadFlag(const std::string &name, bool &fValue) {
 
 bool CBlockTreeDB::LoadBlockIndexGuts()
 {
-    boost::scoped_ptr<leveldb::Iterator> pcursor(NewIterator());
+    std::unique_ptr<CLevelDBIterator> pcursor(NewIterator());
```
dgenr8: Ditto

src/leveldbwrapper.cpp
dagurval: These are removed in a later commit. I just removed them here to avoid having to cherry-pick them away later.

src/txdb.cpp
dagurval: We got unique_ptr with C++11, I prefer to use std when possible. Any reason to use scoped_ptr over unique_ptr?

dgenr8:
ACK

PR #277, opened 2017-11-03
Bitcoin Cash hard fork for DAA and mandatory script flags

dgenr8:
Implement the November 13th, 2017 Bitcoin Cash hard fork as specified at https://github.com/Bitcoin-UAHF/spec/blob/master/nov-13-hardfork-spec.md

The new difficulty adjustment algorithm (DAA) uses a sliding 144-block window and replaces the emergency difficulty adjustment (EDA).

PR #278, opened 2017-11-08
Use Bitcoin Cash network magic

dagurval:
This changes the network magic bytes when the node runs on the Bitcoin Cash network.

These bytes were also used in the database. For backward compatibility, we still use the old byte values for database.

This change is not backward compatible with older nodes. This is fine, since we're doing a hard fork in a few days.

### DIFF
--- a/qa/rpc-tests/test_framework/mininode.py
+++ b/qa/rpc-tests/test_framework/mininode.py
@@ -1498,9 +1498,9 @@ class NodeConn(asyncore.dispatcher):
         b"blocktxn": msg_blocktxn
     }
     MAGIC_BYTES = {
-        "mainnet": b"\xf9\xbe\xb4\xd9",   # mainnet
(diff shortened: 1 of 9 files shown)


dgenr8:
Reviewed, no changes to request.
All unit and RPC tests pass.
Tested successfully together with #277 through the DAA fork on cash testnet.

On mainnet, there are a large number of peers rejected but I do have 8 outbound connections to sensible remote versions.  However 
 - One connection is to a note claiming to run XT G release (is this you @dagurval?)
 - I have no inbound connections

dagurval:
> One connection is to a node claiming to run XT G release (is this you @dagurval?)

That's my node. It has 8 inbound connections. Inbound from these clients: ABC 0.16.0, Classic 1.3.7 and Classic 1.3.8, XT:0.11.0H

dgenr8:
Got inbounds.
ACK

PR #280, opened 2017-11-13
Ensure backupwallet fails if pathDest is set to pathSource

dgenr8:
From https://github.com/bitcoin/bitcoin/pull/11376

Previous behaviour was to destroy the wallet (to zero-length)


dagurval:
utACK

PR #290, opened 2017-12-28
Fixes to smartfees.py

dagurval:
Includes https://github.com/bitcoin/bitcoin/pull/7980

### DIFF
--- a/qa/rpc-tests/smartfees.py
+++ b/qa/rpc-tests/smartfees.py
@@ -7,6 +7,7 @@
 # Test fee estimation code
 #
 
+from collections import OrderedDict
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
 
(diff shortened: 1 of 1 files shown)


qa/rpc-tests/smartfees.py
```diff
@@ -90,7 +90,8 @@ def split_inputs(from_node, txins, txouts, initial_split = False):
     # If this is the initial split we actually need to sign the transaction
     # Otherwise we just need to insert the property ScriptSig
     if (initial_split) :
-        completetx = from_node.signrawtransaction(rawtx)["hex"]
+        completetx = from_node.signrawtransaction(
+                rawtx, None, None, "ALL|FORKID")["hex"]
```
dgenr8: This is the default anyway, and the tests generally don't specify the sighashtype unless a non-default value is being used.

dgenr8: I'm happy to add this, can you just confirm you want it?

dagurval: I'll check if the first patch fixes the issue by itself. If not, then perhaps there is an issue with defaults.

dagurval:
The first commit fixes the issue by itself. Removed the sighashtype change.

PR #294, opened 2018-01-12
Remove warnings on OS X

dagurval:
This removes lots of warnings during compilation on OS X


dgenr8:
ACK on detailed visual check

PR #300, opened 2018-01-17
Return all available information via RPC call "validateaddress"

dagurval:
* Return all available information via RPC call "validateaddress - https://github.com/bitcoin/bitcoin/pull/6262
* Fix crash in validateaddress with -disablewallet - https://github.com/bitcoin/bitcoin/pull/6970


dgenr8:
utACK

PR #305, opened 2018-01-19
Prepare for non-Base58 addresses

dagurval:
Includes changes from:
https://github.com/bitcoin/bitcoin/pull/11247 - Partial cherry-pick to get IsSpendable
https://github.com/bitcoin/bitcoin/pull/11117 - Prepare for non-Base58 addresses
https://github.com/bitcoin/bitcoin/pull/11259 - Remove duplicate destination decoding

This is a prerequisite to supporting cashaddr.


dgenr8:
The pick of https://github.com/bitcoin/bitcoin/pull/11247 is actually complete, not partial.

dagurval:
It doesn't have the one-liner change in `if (!model->IsSpendable)` clause because we don't support custom change addresses in Qt. But that's minuscule, so on second thought, yes I'd call it a complete pick.

dgenr8:
Ok, I missed that.

PR #313, opened 2018-01-24
test: move accounting_tests and rpc_wallet_tests to wallet/test

dagurval:
https://github.com/bitcoin/bitcoin/pull/7905

Prerequisite for cashaddr tests

PR #308, opened 2018-01-23
Kill insecure_random and associated global state

dagurval:
https://github.com/bitcoin/bitcoin/pull/8914

FastRandomContext is used by cashaddr tests

PR #314, opened 2018-01-26
Support cashaddr

dagurval:
Depends on #305, #308, #313

This adds cashaddr support. Needs more testing - marking as WIP for now.


dgenr8:
The dimmed-out example addresses in the send dialog are cashaddrs even when `-usecashaddr=0`, but are network-sensitive.

PR #317, opened 2018-01-30
Bump univalue

dagurval:
Pulls the conversion to subtree from https://github.com/bitcoin/bitcoin/pull/6637, but not the univalue subtree merge.

Rather than pulling over all the subsequent subtree merges from Core, I've pulled over HEAD from upstream (`git remote add univalue https://github.com/bitcoin-core/univalue.git; git subtree add --prefix src/univalue univalue master --squash`)


dgenr8:
`fundrawtransaction.py` failed once with "Insufficient funds" but I've not been able to repeat it.  I'll do more testing for #320.

dgenr8:
Pushed to master using the procedure above with @dagurval as committer (cherry-picker).

PR #293, opened 2018-01-12
Framework for handling double spends

dagurval:
This moves logic for handling double spends out of AcceptToMemoryPool. Each distinct action (relay, notify wallet, logging) is implemented as a separate "respend action". 

Unit test coverage is added for all features.

This also improves logging (enable with debug=respend) and fixes an issue where notification to wallet would be stopped by the relay limiter. The relay filter is replaced with a rolling bloom filter.

This will make it easier to add additional features such as https://github.com/bitcoinxt/bitcoinxt/issues/246 or RPC notifications on respends. New features can be added by implementing the _RespendAction_ interface.

PR #322, opened 2018-02-06
Improve readability in ATMP

dgenr8:
Thanks to #293 it is finally possible to stop hijacking the final `return true` in ATMP.

The diff looks less scary with `git show --diff-algorithm=histogram`.


dagurval:
Nice. 

Also less scary with [?w=1](https://github.com/bitcoinxt/bitcoinxt/pull/322/files?w=1) :-)

dgenr8:
whaa.... wow!

PR #327, opened 2018-02-09
Make mempool entry available to respend actions

dgenr8:
Enables respend actions to make use of pool entry time and various calculated quantities.


dgenr8:
Update ... I'm changing to use a txiter instead of CTxMemPoolEntry& to allow a tx1 set to be built efficiently (this is useful in respend analysis).

dagurval:
Good idea. utACK

dgenr8:
Squashed.

PR #328, opened 2018-02-12
Use respend framework for respends found in blocks

dgenr8:
Reuse the new framework and avoid the need to repeat logging in CTxMemPool::removeConflicts(...)

Functional test `txn_doublespend.py` should succeed and log the respend found in a block, while `txn_clone.py` should succeed and not log the respend because the RespendLogger doesn't log equivalent clones.

### DIFF
--- a/src/respend/mempoolremover.cpp
+++ b/src/respend/mempoolremover.cpp
@@ -1,0 +1,36 @@
+// Copyright (c) 2018 The Bitcoin developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "respend/mempoolremover.h"
+
+namespace respend {
+
+MempoolRemover::MempoolRemover(CTxMemPool& pool, std::list<CTransaction>& removed) :
+        pool(pool), removed(removed), valid(false)
+{
+}
+
+bool MempoolRemover::AddOutpointConflict(
+        const COutPoint&, const CTxMemPool::txiter mempoolEntry,
+        const CTransaction& respendTx,
+        bool seenBefore, bool isEquivalent)
+{
+    tx1s.insert(mempoolEntry);
+
+    // Keep gathering conflicting transactions
+    return true;
+}
+
+bool MempoolRemover::IsInteresting() const {
+    return true;
+}
+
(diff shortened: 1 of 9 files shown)


src/respend/mempoolremover.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+bool MempoolRemover::AddOutpointConflict(
+        const COutPoint&, const CTxMemPool::txiter mempoolEntry,
+        const CTransaction& respendTx,
+        bool seenBefore, bool isEquivalent)
+{
+    tx1s.insert(mempoolEntry);
+
+    // Keep gathering conflicting transactions
+    return true;
```
dagurval: `return true` will make it to look for more conflicting inputs in the same transactions, so you probably want to return false here

src/respend/mempoolremover.h
```diff
@@ -0,0 +1,36 @@
...
+        void SetValid(bool v) override {
+            valid = v;
+        }
+        void Trigger() override;
+
+    private:
+        CTxMemPool& pool;
+        std::list<CTransaction>& removed;
+        CTxMemPool::setEntries tx1s;
+        bool valid;
```
dagurval: `valid` isn't used for anything, how about removing it and make `SetValid` throw when called as `SetValid(false)`?

src/respend/respenddetector.cpp
```diff
@@ -77,7 +77,7 @@ void RespendDetector::CheckForRespend(
         conflictingOutpoints.push_back(outpoint);
 
         CTxMemPool::txiter poolIter = pool.mapTx.find(spendIter->second.ptx->GetHash());
-        if (poolIter == pool.mapTx.end())
+        if (poolIter == pool.mapTx.end() || poolIter->GetTx() == tx)
```
dagurval: A unittest for this case would be nice (but not a blocker for this PR)

dagurval:
Looks good. Nice to see duplicate mempool logic removed. Had a couple of nits.

dgenr8:
Originally I didn't have a `valid` flag in `MempoolRemover` but because `SetValid` is required to be implemented and RespendLogger doesn't do anything unless `valid`, I added it with the intention of requiring it.  Update pushed.

PR #335, opened 2018-02-14
Enter safe mode and warn of network upgrade

dagurval:
If we plan a release before implementing the changes in next [scheduled network upgrade](https://github.com/bitcoincashorg/spec/pull/53), this will make the node enter safe mode and warn the user of outdated software.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4425,6 +4425,19 @@ string GetWarnings(string strFor)
         strStatusBar = strRPC = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
     }
 
+    {
+        LOCK(cs_main);
+        CBlockIndex* tip = chainActive.Tip();
+        if (tip && IsMay2018HFActive(tip->GetMedianTimePast()))
+        {
+            std::string err =
+                "Error: This version of Bitcoin XT is outdated, please upgrade. "
+                "Sending transactions MAY CAUSE LOSS OF FUNDS. "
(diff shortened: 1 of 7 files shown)


dgenr8:
I fear that by triggering on an expected fork date defined by a committee, we expose ourselves to the same problem as people who actually plan a fork with an unrealistic date -- namely scrambling to un-do the warning when the date has to be changed.

Could we instead trigger safe mode when no blocks are seen for 130 minutes (the low-side trigger in #147)?

dagurval:
We know that there is likely to be a majority fork at that time. So far, it doesn't look like it will be contentious. We don't know if there will be a minority fork. 

If we assume that minority fork has low hash power and/or luck, then the _no block seen_ check will work.

> [..] we expose ourselves to scrambling to un-do the warning when the date has to be changed.

No need to scramble. The nodes won't be crippled. Subset of the node functionality is locked out, mainly wallet functions. 

The node operator will have to assess the network situation and either `-disablesafemode` or upgrade to a version without the warning. The point is to not let the users unknowingly operate on the wrong side of the chain (_wrong_ being up to them to decide).

The warning could use some rewording.

dgenr8:
`CheckForkWarningConditions() ` also triggers safe mode when a ~6 block valid or invalid fork is found, and it doesn't depend on rate of new valid blocks.  Hard-coding the fork MTP in the non-fork client only gets us about an hour of extra warning.

dagurval:
That seems like a better strategy. Closing this.

PR #336, opened 2018-02-15
Introduce DummyArgGetter

dagurval:
This removes the need to implement a custom DummyArgGetter when overriding options in unittests.

Also turned Opt into a class (instead of struct)


dagurval:
rebased

dgenr8:
This is nice!

PR #343, opened 2018-02-22
Improve handling of block announcements

dagurval:
The first commit is a bugfix where we would exceed MAX_BLOCKS_IN_TRANSIT_PER_PEER. The second commit mitigates having us misbehave nodes for sending unconnected headers. See commit description for details.

### DIFF
--- a/src/blockannounce.cpp
+++ b/src/blockannounce.cpp
@@ -57,13 +57,18 @@ BlockAnnounceReceiver::DownloadStrategy BlockAnnounceReceiver::pickDownloadStrat
     if (hasBlockData())
         return DONT_DOWNL;
 
+    NodeStatePtr nodestate(from.id);
+    // Peer has an unresolved unconnecting header situation, requesting new
+    // blocks may cause yet another unconnecting header to arrive.
+    if (nodestate->unconnectingHeaders)
+        return DONT_DOWNL;
+
+    if (nodestate->nBlocksInFlight >= MAX_BLOCKS_IN_TRANSIT_PER_PEER)
+        return DONT_DOWNL;
+
     if (!fetchAsThin()) {
 
         if (blocksInFlight.isInFlight(block))
-            return DONT_DOWNL;
-
-        NodeStatePtr nodestate(from.id);
-        if (nodestate->nBlocksInFlight >= MAX_BLOCKS_IN_TRANSIT_PER_PEER)
             return DONT_DOWNL;
 
         if (Opt().AvoidFullBlocks()) {
--- a/src/blockannounce.cpp
+++ b/src/blockannounce.cpp
@@ -176,12 +176,52 @@ bool blocksConnect(const std::vector<uint256>& blocksToAnnounce) {
     return true;
 }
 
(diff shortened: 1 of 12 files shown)


src/test/processmessage_tests.cpp
```diff
@@ -82,7 +82,7 @@ struct DummyXThinProcessor : public XThinBlockProcessor {
         BlockHeaderProcessor& h) : XThinBlockProcessor(f, w, h), misbehaved(0)
     { }
 
-    virtual void misbehave(int howmuch) {
```
dgenr8: Ah so THAT's what the override kwd is good for!

src/blockannounce.cpp
```diff
@@ -176,14 +176,54 @@ bool blocksConnect(const std::vector<uint256>& blocksToAnnounce) {
...
 bool BlockAnnounceSender::canAnnounceWithHeaders() const {
 
+    if (to.blocksToAnnounce.empty())
+        return false;
+
     // Set too many blocks to announce. In this corner case
     // we revert to announcing blocks as inv even though
     // peer prefers header announcements.
     if (to.blocksToAnnounce.size() > MAX_BLOCKS_TO_ANNOUNCE)
         return false;
```
dgenr8: Why not limit how many are actually announced, instead of reverting to inv?

src/main.cpp
```diff
@@ -320,6 +320,10 @@ struct MarkBlockAsInFlight : public BlockInFlightMarker {
         nQueuedValidatedHeaders += newentry.fValidatedHeaders;
         list<QueuedBlock>::iterator it = state->vBlocksInFlight.insert(state->vBlocksInFlight.end(), newentry);
         state->nBlocksInFlight++;
+        if (state->nBlocksInFlight > MAX_BLOCKS_IN_TRANSIT_PER_PEER) {
```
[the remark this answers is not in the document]
dagurval: Correct. I don't want it to pass silently if we regress again. Throwing could be alternative to logging.

src/compactblockprocessor.cpp
```diff
@@ -29,7 +29,11 @@ void CompactBlockProcessor::operator()(CDataStream& vRecv, const CTxMemPool& mem
         return;
     }
 
-    if (requestConnectHeaders(block.header)) {
+    // We (may) have requested node to announce blocks to us and it will
+    // continue to do so even if we haven't kept up. Don't misbehave it for
+    // sending us unconnected headers.
+    bool bumpUnconnecting = false;
```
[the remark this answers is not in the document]
dagurval: The misbehave is primarily meant to hinder us from looping like:
received headers -> headers missing, request headers -> received headers -> headers missing, request headers -> ...

src/blockannounce.cpp
dagurval: We want to announce the chain tip and in this case we can't do it with headers

src/test/processmessage_tests.cpp
dagurval: `override` has saved me from making errors like this so many times... but no this time :cry:

PR #345, opened 2018-03-02
Compact blocks updates

dagurval:
This bumps the upper boundry of the index range from 2^16 to 2^32.

Also includes performance improvements for reconstructing thin blocks.

Depends in #343


dagurval:
Comment updated. Good catch @dgenr8

PR #346, opened 2018-03-02
add acceptnonstdtxn flag

gandrewstone:
This flag (when set to 0) makes testnet and regtest behave like mainnet by rejecting nonstandard transactions.  The flag name and behavior is awkwardly chosen to be compatible with other Satoshi clients.

Also I fixed a small issue in the description of datacarriersize.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -902,6 +902,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             return InitError(strprintf(_("Invalid amount for -minrelaytxfee=<amount>: '%s'"), mapArgs["-minrelaytxfee"]));
     }
 
+    bool fStandard = !GetBoolArg("-acceptnonstdtxn", !Params().RequireStandard());
+    // If we specified an override but that override was not accepted then its an error
+    if (fStandard != Params().RequireStandard())
+        return InitError(
+            strprintf("acceptnonstdtxn is not currently supported for %s chain", chainparams.NetworkIDString()));
+
 #ifdef ENABLE_WALLET
     if (mapArgs.count("-mintxfee"))
     {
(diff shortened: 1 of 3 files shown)


src/init.cpp
```diff
@@ -902,6 +902,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             return InitError(strprintf(_("Invalid amount for -minrelaytxfee=<amount>: '%s'"), mapArgs["-minrelaytxfee"]));
     }
 
+    bool fStandard = !GetBoolArg("-acceptnonstdtxn", !Params().RequireStandard());
+    // If we specified an override but that override was not accepted then its an error
+    if (fStandard != Params().RequireStandard())
```
dgenr8: This doesn't work in the case of regtest, where `-acceptnonstdtxn=false` should be accepted.  Use `if (Params().RequireStandard() && !fStandard)` here?

src/init.cpp
```diff
@@ -902,6 +902,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+        return InitError(
+            strprintf("acceptnonstdtxn is not currently supported for %s chain", chainparams.NetworkIDString()));
```
dgenr8: Prefix parameter name with '-'.

src/init.cpp
gandrewstone: Its accepted.  I tried it, but also here's how the logic works:

Params().RequireStandard() 
=> fRequireStandard || !GetBoolArg("-acceptnonstdtxn", true); (def of fn)
=> !GetBoolArg("-acceptnonstdtxn", true); (fRequireStandard == false)
=> true (because acceptnonstdtxn=false was an arg)

bool fStandard = !GetBoolArg("-acceptnonstdtxn", !Params().RequireStandard());
= true (because acceptnonstdtxn=false was an arg)

if (fStandard != Params().RequireStandard()) =>
if (true != true)

dgenr8: Ok, it works because the parameter being validated is already affecting the behavior of CChainParams::RequireStandard().  I guess making CChainParams:fRequireStandard public would be worse.

dgenr8:
Thanks @gandrewstone for this change, and for cash interop!

PR #349, opened 2018-03-05
Fix OOM when deserializing UTXO entries with invalid length

dagurval:
https://github.com/bitcoin/bitcoin/pull/7933


dgenr8:
ACK

PR #351, opened 2018-03-07
[qa] Don't manually sync blocks in p2p-compactblocks

dagurval:
The network is not segregated by segregated witness, so no need to manually sync it.


dgenr8:
utACK

PR #368, opened 2018-04-04
Add missing lock in getblocktxn

dagurval:
I have not been able to reproduce the pruning.py failures we're having after applying this fix.

Also adds some logging that helped me find the problem.


dgenr8:
utACK

PR #370, opened 2018-04-04
Prepare for re-enabled opcodes

dagurval:
Prepare for re-enabled opcodes for the May 2018 protocol upgrade - https://reviews.bitcoinabc.org/D1141
Add a flag to gate opcodes activation for the new HF - https://reviews.bitcoinabc.org/D1208
Add helper function for disabled opcodes - https://reviews.bitcoinabc.org/D1209


src/core_read.cpp
```diff
@@ -30,9 +30,9 @@ CScript ParseScript(std::string s)
 
     if (mapOpNames.empty())
     {
-        for (int op = 0; op <= OP_NOP10; op++)
+        for (int op = 0; op <= FIRST_UNDEFINED_OP_VALUE; op++)
```
dgenr8: Why <= rather than <?  This changes the semantics.

Danconnolly: Yes, I'm not sure how this happened. In https://reviews.bitcoinabc.org/D1141 it is "<".

dagurval: Good eye @dgenr8. 

That’s a bad merge by me. ABC reformatted all the files so cherry picking involves a lot of manual merging.

dagurval:
Fixed issue found by @dgenr8

Issue #323, opened 2018-02-06
Tests TooLargeVerack and TooLargePing are testing uninitialized data


dagurval:
Tests [_TooLargeVerack_ and _TooLargePing_](https://github.com/bitcoinxt/bitcoinxt/blob/master/src/test/ReceiveMsgBytes_tests.cpp#L99) create a dummy node, but don't insert it into NodeStatePtr.

This causes `GetNodeStateStats` to return immediately without initializing `stats`. The test does checks on uninitialized data.

Found when debugging why these tests randomly fail.

From gdb:
```
ReceiveMsgBytes_tests::TooLargeVerack::test_method (this=0x7fffffffc4b0)                              
    at test/ReceiveMsgBytes_tests.cpp:119          
119         BOOST_CHECK(stats.nMisbehavior > 0);   
(gdb) p stats            
$2 = {nMisbehavior = 1446951336, nSyncHeight = 21845, nCommonHeight = 1440140288,                     
  vHeightInFlight = std::vector of length 0, capacity 0}  
```

dagurval:
Fixed in https://github.com/bitcoinxt/bitcoinxt/commit/dc52e1438ba3391f39e3c266942c30df76914cf5

PR #386, opened 2018-04-13
Don't require CASH service bit for outgoing connections

### LINKED ISSUE #309: Don't require CASH service bit for outgoing connections
BU is removing the service bit `NODE_BITCOIN_CASH`, relying only on the network magic in https://github.com/BitcoinUnlimited/BitcoinUnlimited/pull/905/commits/a92953f0d219fd7e89c289409c9a940fa5f4eda9

We require this flag for outgoing connections. This means that XT will no longer make outgoing connections to Bitcoin Unlimited nodes.

We need to remove this requirement, preferably before the May network upgrade. We should also sunset announcing the NODE_BITCOIN_CASH service bit. Perhaps one year from now.

dagurval:
Fixes #309

### DIFF
--- a/src/test/p2p_protocol_tests.cpp
+++ b/src/test/p2p_protocol_tests.cpp
@@ -6,6 +6,7 @@
 #include "net.h"
 
 #include "test/test_bitcoin.h"
+#include "test/thinblockutil.h"
 
 #include <boost/test/unit_test.hpp>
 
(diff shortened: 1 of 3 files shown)


dgenr8:
@dagurval How did you observe the test failure that the second commit fixes?

dagurval:
The tests aren't very good. They just test if "ProcessMessage" returns false, not what caused it to return false. In this case, they were returning false because the dummy nodes were _outgoing_ and _not signalling cash_. After this fix, this condition didn't trigger, so they took the original path again, the one that uses uninitialized nodestates.

Why this worked before I don't know. I fixed the same thing here: https://github.com/bitcoinxt/bitcoinxt/issues/323

PR #389, opened 2018-04-17
[http] Scale request accept limit with block size

dagurval:
This is my take on https://reviews.bitcoinabc.org/D1285


dgenr8:
QA tests are sometimes returning this error which Jenkins also got
`Unexpected exception caught during testing: bitcoind exited with status -11 during initialization`

dagurval:
`GetMaxBlockSize` was missing a check for `nullptr`. Fixed.

@dgenr8 There is no way to set the global MOCKTIME to a specific value.

dgenr8:
ACK, though I think having "insecure" in the name is overkill.  I mean, it's not `TipInsecure()` ...

PR #392, opened 2018-04-25
Add SizeEstimate to CDBBatch

dagurval:
e66dbde6d from Core#10195


dgenr8:
utACK Noted the fixup required because we don't have obfuscation.

PR #393, opened 2018-04-25
error in disconnect for disk corruption, not inconsistency

dagurval:
f54580e7e from Core#10195


dgenr8:
utACK. Noted ApplyTxInUndo made static rather than changing the return type in the test version.

PR #394, opened 2018-04-25
Introduce CHashVerifier + specialization of SipHash for 256 + 32 bit data

dagurval:
Avoid VLA in hash.h - https://github.com/bitcoin/bitcoin/pull/9791
e484652fc and e484652fc from Core#10195


dgenr8:
Should we apply the tx version 2 bump before this?

dagurval:
I don't see why we need to, but I've pulled over tx version 2 (https://github.com/bitcoinxt/bitcoinxt/pull/405), so go ahead :-)

dgenr8:
You're right, I was noticing the nearby differences in `hash_tests.cpp` but the changes are independent.

PR #395, opened 2018-04-25
Divergence between 32- and 64-bit when hashing >4GB affects `gettxoutsetinfo`

dagurval:
https://github.com/bitcoin/bitcoin/pull/7848

Skipped the release-notes commit when cherry-picking.


dgenr8:
utACK

PR #396, opened 2018-04-25
Remove/ignore tx version in utxo and undo 

dagurval:
d34242430 from Core#10195, in addition to a change on BIP64 where tx version is replaced with dummy version 0.

On top of #395  and #393 

No conflicts.


dgenr8:
Rebased-as-@dagurval

dgenr8:
utACK

PR #397, opened 2018-04-26
Report on-disk size in gettxoutsetinfo

dagurval:
On top of #392 and #396.

c3aa0c119 from Core#10195

Trivial merge.


dgenr8:
Rebased-as-@dagurval

dgenr8:
utACK

PR #401, opened 2018-04-26
Replace BCC ticker with BCH in debian package

### LINKED ISSUE #383: Debian installer uses ticker BCC
The [debian installer](https://github.com/bitcoinxt/bitcoinxt/tree/master/contrib/gitian-debian) still uses ticker BCC. Needs to be updated to BCH.

dagurval:
The package in our Debian repository already has this patch applied.

fixes #383


dgenr8:
utACK

PR #404, opened 2018-04-27
bitcoin-tx JSON tests + test util improvements

dagurval:
On top of #403

https://github.com/bitcoin/bitcoin/pull/8829 - Add bitcoin-tx JSON tests
https://github.com/bitcoin/bitcoin/pull/8830 - [test] Add option to run bitcoin-util-test.py manually
https://github.com/bitcoin/bitcoin/pull/8836 - bitcoin-util-test.py should fail if the output file is empty
https://github.com/bitcoin/bitcoin/pull/8837 - allow bitcoin-tx to parse partial transactions _[partial pick]_
https://github.com/bitcoin/bitcoin/pull/8881 - Add some verbose logging to bitcoin-util-test.py
https://github.com/bitcoin/bitcoin/pull/9032 - test: Add format-dependent comparison to bctest
https://github.com/bitcoin/bitcoin/pull/9023 - Add logging to bitcoin-util-test.py
https://github.com/bitcoin/bitcoin/pull/9069 - Clean up bctest.py and bitcoin-util-test.py
https://github.com/bitcoin/bitcoin/pull/9945 - Improve logging in bctest.py if there is a formatting mismatch

Trivial cherry picks. Final commit updates json files to replaces legacy addresses with cashaddr.

PR #405, opened 2018-04-27
Bump transaction version default to 2 

dagurval:
On top of #404 

https://github.com/bitcoin/bitcoin/pull/7562 - Bump transaction version default to 2                  
https://github.com/bitcoin/bitcoin/pull/9376 - Remove unused test files and references                
https://github.com/bitcoin/bitcoin/pull/9406 - Re-enable a blank v1 Tx JSON test

* Skipped changes in src/test/txvalidationcache_tests.cpp, because we don't have it.
* Removed "hash" from data files
* Replaced legacy addresses with cashaddr in data files

Commit "tiny test fix for mempool_tests (c5d746a)" skipped, it breaks the test. Are ties not broken by hash @dgenr8?


dgenr8:
Rebased-as-@dagurval

dgenr8:
@dagurval We [don't](https://github.com/bitcoinxt/bitcoinxt/blob/master/src/txmempool.h#L252) break ties by hash in the ancestor index (it could be added last, after Time).  Nevertheless our `mempool_tests.cpp` is deterministic because tx6 is added last and expected to sort last, and tx7 doesn't require a tiebreak.

PR #406, opened 2018-04-28
Correct Windows builds using the wrong FD_SETSIZE

dgenr8:
https://github.com/BitcoinUnlimited/BitcoinUnlimited/pull/1047

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -7,6 +7,9 @@
 #include "config/bitcoin-config.h"
 #endif
 
+// compat.h must be included first to correctly define FD_SETSIZE
+#include "compat.h"
+
 #include "net.h"
 #include "addrman.h"
 #include "chainparams.h"
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -7,6 +7,9 @@
 #include "config/bitcoin-config.h"
 #endif
 
+// compat.h must be included first to correctly define FD_SETSIZE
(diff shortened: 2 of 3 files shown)


src/util.cpp
```diff
@@ -671,7 +673,7 @@ bool TruncateFile(FILE *file, unsigned int length) {
  */
 int RaiseFileDescriptorLimit(int nMinFD) {
 #if defined(WIN32)
-    return 2048;
```
dagurval: So we don't break this again, how about:
```
static_assert(FD_SETSIZE > 64, "compat.h needs to be included first, see PR#406");
```
?

src/net.cpp
```diff
@@ -7,6 +7,10 @@
 #include "config/bitcoin-config.h"
 #endif
 
+// must include first to ensure FD_SETSIZE is correctly set
+// otherwise the default of 64 may be used on Windows
```
dagurval: Suggest rewording to 
```
// compat.h must be included first to correctly define FD_SETSIZE
```
This is more explicit on what must be included, less unnecessarily details. Details about overriding the value 64 can go into compat if we care. Also, 64 is only default on mingw, not MSVC.

dgenr8:
Pushed changes suggested by @dagurval.

dagurval:
Thanks!

PR #407, opened 2018-04-30
Tx metadata in all undo records & new coins interface

dagurval:
422634e2f and cb2c7fdac from Core#10195

No conflicts.

### DIFF
--- a/src/coins.h
+++ b/src/coins.h
@@ -19,6 +19,68 @@
 #include <boost/foreach.hpp>
 #include <unordered_map>
 
+/**
+ * A UTXO entry.
+ *
+ * Serialized format:
+ * - VARINT((coinbase ? 1 : 0) | (height << 1))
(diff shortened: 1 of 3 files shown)


dgenr8:
This change commits (or maybe further commits) us to a chainstate rebuild on next launch.  Just making sure we see a path to the complete upgrade from here.

dagurval:
This is backward compatible, but not forward compatible. It won't be possible to downgrade without reindexing.

Backward compatibility is also maintained, see `ApplyTxInUndo` in #411. Although, I don't mind  simplifying the code by removing backward compatibility, thus forcing reindex. Computer time is less valuable than developer time.

dgenr8:
My only concern is that we have the complete upgrade in sight when we start sacrificing compatibility.

dgenr8:
utACK

PR #409, opened 2018-04-30
Use coin in undo

dagurval:
7d991b55d and 422634e2f from Core#10195

On top of #407 

Trivial conflict in main (function signature)

Changes in coins_tests skipped. I plan to introduce these changes after a later patch (05293f3cb removes ModifyNewCoins, that we never introduced).

The change in undo.h introduced `MAX_INPUTS_PER_BLOCK` that does not make sense for us. This constant is replaced with MAX_INPUTS_PER_TX, [as done in ABC](https://reviews.bitcoinabc.org/rABC421bf43c7b8af31cb16fa85818e45e1c0833ae25).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1633,7 +1633,7 @@ enum DisconnectResult
  * @param out The out point that corresponds to the tx input.
  * @return A DisconnectResult as an int
  */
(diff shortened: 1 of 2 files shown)


dgenr8:
Rebased as @dagurval.

dgenr8:
ACK. Tested with 1000-block reorg.

PR #410, opened 2018-05-02
Introduce new per-txout CCoinsViewCache functions

dagurval:
On top of #407 

Constant MAX_OUTPUTS_PER_BLOCK replaced with MAX_OUTPUTS_PER_TX, [as done in ABC](https://reviews.bitcoinabc.org/rABCf701ed902693028c2c198592cc77bd4e950064db)

PR #408, opened 2018-04-30
Benchmark cherries

dagurval:
https://github.com/bitcoin/bitcoin/pull/8039 - bench: Add crypto hash benchmarks
https://github.com/bitcoin/bitcoin/pull/8107 - bench: Added base58 encoding/decoding benchmarks
https://github.com/bitcoin/bitcoin/pull/8111 - Benchmark SipHash
https://github.com/bitcoin/bitcoin/pull/8115 - Avoid integer division in the benchmark inner-most loop.
https://github.com/bitcoin/bitcoin/pull/8873 - Add microbenchmarks to profile more code paths.
https://github.com/bitcoin/bitcoin/pull/9200 - bench: Fix subtle counting issue when rescaling iteration count

Most of these applied with no or trivial conflicts, except for https://github.com/bitcoin/bitcoin/pull/8873. The modifications needed were done in a separate commit.

PR #411, opened 2018-05-02
Switch from per-tx to per-txout CCoinsViewCache methods in some places

dagurval:
f68cdfe92 from Core#10195

On top of #410 and #408 

Mostly trivial merges, notable conflicts:
* tx_verify.cpp - patched to keep error() logging
* sigopcount_tests.cpp - skipped changes for a test we don't have


dgenr8:
ACK

PR #414, opened 2018-05-03
Switch CScriptCheck to use Coin instead of CCoins

dagurval:
On top of #411 

c87b957a3 and 8b3868c1b from Core#10195

These conflicted on almost every code line, however the conflicts were trivial.


dgenr8:
urACK

PR #415, opened 2018-05-03
Update coins_tests.cpp as of 961e4839

dagurval:
On top of #410 

This copies over coins_tests.cpp at Bitcoin Core commit
"961e4839 - Switch tests from ModifyCoins to AddCoin/SpendCoin".

Reason for updating the tests by copy is because the updates were
written for an intermediate solution "ModifyNewCoins", we never
introduced this solution, nor do we want to.

No conflicts.


dgenr8:
ACK

PR #422, opened 2018-05-12
fix build on osx

jeffgran:
Hello, I'm new here. 👋 

Wanting to get my hands dirty and help out a little. The first problem I ran into was that I couldn't build the project in my OSX environment, so I took to fixing that. This is the result.

Specifically

- OSX needs `libevent` as a dependency (probably other envs do too?) but it wasn't in the build doc so I added that.
- boost as of v 1.66+ has some differently typed function signatures, I guess. `brew` on osx installs 1.67.0 as of this writing. So I fixed the incompatibility, same as was done in core [here](https://github.com/bitcoin/bitcoin/pull/11847/files)
- similarly, qt 5.7+ (`brew` installs 5.10.1 as the latest) had compatibility issues. Core fixed that with [this change](https://github.com/bitcoin/bitcoin/pull/9169), ~~but I tried that and it didn't work for me. Looking deeper, their fix seems to be a more complex fix than the simplest fix, which was intended to be "future proof" for newer compilers, but based on the fact that it didn't work for me here in the future that wasn't right. So I just have the simple fix here in this PR (just adding the `-std=c++11` to `OBJCXXFLAGS` because for whatever reason qt doesn't respect the flag in `CXXCPP` and/or `CXX` set by `AX_CXX_COMPILE_STDCXX`.~~ and that is replicated here.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -337,7 +337,7 @@ case $host in
 
      AX_CHECK_LINK_FLAG([[-Wl,-headerpad_max_install_names]], [LDFLAGS="$LDFLAGS -Wl,-headerpad_max_install_names"])
      CPPFLAGS="$CPPFLAGS -DMAC_OSX"
-     OBJCXXFLAGS="$CXXFLAGS"
+     OBJCXXFLAGS="-std=c++11 $CXXFLAGS"
      ;;
    *linux*)
      TARGET_OS=linux
--- a/configure.ac
+++ b/configure.ac
@@ -40,7 +40,6 @@ else
   CXXFLAGS_overridden=no
 fi
 AC_PROG_CXX
-m4_ifdef([AC_PROG_OBJCXX],[AC_PROG_OBJCXX])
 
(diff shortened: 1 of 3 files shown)


dgenr8:
Thanks @jeffgran.  It would be great to get the OSX build working well (and binaries too).
Ping @kyuupichan

dagurval:
@xt-bot ok to test

jeffgran:
Thanks for the prod @dagurval. I looked again more carefully and I was indeed able to get Core's fix to work. I'm working part-way on intuition here but I guess the part where we assign `$CXX` to `$OBJCXX` has to come *after* the autoconf directive `AX_CXX_COMPILE_STDCXX([11], [noext], [mandatory])` since that is probably what sets up those env vars in the first place. 

Anyway I pushed another commit and I think this is cleaner now.

dagurval:
Great 👍

ACK, tested compilation on OSX.

PR #423, opened 2018-05-14
read config file before checking for existence of datadir

### LINKED ISSUE #241: Hang at startup if -datadir does not exist.
### Description:
Bitcoin XT is unable to start if parameter _datadir_ is set, but directory does not exist.

The process hangs. When killing the process an exception error is shown:
```
$ ./src/bitcoind -reindex

^C

************************
EXCEPTION: N5boost10filesystem16filesystem_errorE       
boost::filesystem::space: Operation not permitted       
bitcoin in AppInit()  
```

### How to reproduce

Add `datadir=/path/does/not/exist` to bitcoin.conf. Start Bitcoin XT.

### Expected behavior

Log message about datadir not existing and exit

jeffgran:
Fixes #241

The code was checking for the existence of the `datadir` *before* reading and processing the config file. This means that when the datadir was only specified in the config file, we ended up checking whether the default dir existed, which it does, and then reading in a nonexistent datadir after that, causing later code (which assumes this check has already been performed and passed) to have erratic behavior.

~~So the fix is just to move the "read config" code up higher so it happens first. I think in general reading the config file should be as high as possible in this function, or maybe even up higher in the call tree? Not sure how the GUI mode interacts with the conf file yet though.~~

Fixed by checking that the datadir exists inside the `ReadConfigFile` function, just before returning. Cherry-picked from https://github.com/bitcoin/bitcoin/pull/11829

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -592,6 +592,9 @@ void ReadConfigFile(map<string, string>& mapSettingsRet,
     }
     // If datadir is changed in .conf file:
     ClearDatadirCache();
+    if (!boost::filesystem::is_directory(GetDataDir(false))) {
+        throw std::runtime_error(strprintf("specified data directory \"%s\" does not exist.", GetArg("-datadir", "").c_str()));
+    }
 }
 
 #ifndef WIN32


dgenr8:
@jeffgran It looks like Core addressed this issue in https://github.com/bitcoin/bitcoin/pull/11829.

The core fix has the advantages of exiting sooner and prefixing the error message (via thrown exception) with "Error reading configuration file:".  What do you think of changing your branch into a cherry-pick of the core change?  (another advantage is making future cherry-picks easier.)

dagurval:
Thanks for debugging this. I agree that @dgenr8's suggestion to move the check into ReadConfigFile and throw to get the error prefix is a slightly better.

dagurval:
@xt-bot add to whitelist

jeffgran:
Just getting back to this. Thanks for the comments.

I disagree that their fix is better -- it duplicates the check that already exists in the AppInit into another place, and that second place is a function named `ReadConfigFile` -- I think it's strange to check the validity of just this one config setting inside the read-config subroutine. 

However, to argue against myself, I do take your point that cherry-picking core's commits makes it easier to cherry pick others later, and this obviously isn't a super important change so I'll update this to take core's fix a little later.

dgenr8:
@jeffgran You make a good argument. Thank you!

PR #424, opened 2018-05-16
[Tests] Correct tests to pre-thirdfork env

dgenr8:
Once the fork deployed, these tests were sensitive to the mempool being cleared as the fork was "rolled back" by invalidations.


dgenr8:
Added fix for pruning.py.

dagurval:
utACK. What's the reason for pushing the fork into the future, rather than burying it far into the past? It would make more sense to have the tests run on current consensus rules.

dgenr8:
Just fixing the build by restoring the previous test environment.  Better tests welcome.

PR #431, opened 2018-05-22
Move scheduler/threadGroup into common-init instead of per-app

dgenr8:
https://github.com/bitcoin/bitcoin/pull/12266

Resolves shutdown crash during startup caused by scheduler thread not being interrupted and joined before destruction in the case where initialization did not complete.


dagurval:
utACK

PR #435, opened 2018-05-25
At shutdown, join threads before destroying Connman

dgenr8:
This fixes intermittent shutdown crashes due to mutexes that were being destroyed while locked. Tested:
 - bitcoin-qt exits cleanly
 - Previous scenarios showing intermittent crashes crash no longer
 - Fast shutdown for missing rpc credentials does not crash

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -159,7 +159,6 @@ void Interrupt()
     InterruptHTTPRPC();
     InterruptRPC();
     InterruptREST();
-    threadGroup.interrupt_all();
 }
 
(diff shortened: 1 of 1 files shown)


dagurval:
LGTM, I'm curious though if https://github.com/bitcoin/bitcoin/pull/9289 (picked in #436) touches this issue.

dgenr8:
I certainly hope #436 makes this PR unnecessary since it goes to all kinds of trouble to make it safe to delete CConnman before joining the processing threads.

PR #438, opened 2018-05-30
OSX app build updates

### LINKED ISSUE #437: BitCoin-XT Mac version just used the bundle identifier from the official bitcoin client
please change the
CFBundleIdentifier
in 
share/qt/Info.plist.in

from 
org.bitcoinfoundation.Bitcoin-Qt
to something else

each Mac app has to have a unique bundle ID, so stealing the bundle ID from some other project is a super bad idea

jeffgran:
Closes #437 

Cherry-picking a couple of commits from BitcoinABC where they updated their OSX build to use a different identifier, and use the `bitcoincash:` url scheme instead of `bitcoin:`.

#### TODO: 

- [ ] Need to build and test to make sure this works.


jeffgran:
So the docs point at https://github.com/bitcoinxt/bitcoinxt/blob/master/doc/release-process.md for how to build the OSX application bundle. But I don't think I have the powers (i.e. the keys/access) to follow that guide. It talks about building for official releases using gitian, etc. Not sure I can do much else on this... but happy to help test on OSX if needed.

@core-code I am curious how you even found this. We have not built/distributed any bitcoinXT OSX application bundles, have we?

core-code:
you do seem to have at least one official Mac binary release here:
https://github.com/bitcoinxt/bitcoinxt/releases/download/v0.11E/BitcoinXT-0.11E.dmg

i noticed the problem because we have been working on a mac app that can update all your mac apps ( https://www.corecode.io/macupdater/index.html ).
while this works fine fine for more than 20.000 apps, bitcoinxt is one of the handful that messes up our database because it uses the same 'bundle identifier' as another app.

i have workarounds in our database but i wanted to give you an heads-up because there can be many unexpected bugs if you have the same bundle-id as another app e.g. because the preferences will be shared with all apps with the same bundle identifier.

dgenr8:
@core-code Is that old release E .dmg the one your were trying to include? I'm asking because pretty much anyone interested in XT today can't run it (it doesn't have bitcoin cash support and is also simply outdated).  What version have you included via your workarounds? (Should be release I)

core-code:
it seems to be from 2015, 8 of our users still have it installed:

'org.bitcoinfoundation.Bitcoin-Qt'
'0.11.0, Copyright Â© 2009-2015 The Bitcoin Core developers'
'0.11.0'
'0.11.0'
'2015-06-01'
'/Applications/Bitcoin-XT.app'

dgenr8:
@jeffgran It would be most valuable to get the OSX binary gitian build working.  A lot of the process document has to do with comparing fingerprints of builds done by different people to make sure the build is deterministic, and final signing of the binary (which I haven't been doing for the Win build - we currently rely on the signed checksums file).

jeffgran:
@dgenr8 Cool, thanks for the direction! I took a crack at getting the gitian build working. I got pretty close, I got it to build for the most part, but then got deeeeep into the weeds trying to figure out why it crashed near the end of the build script, and found out that it seems to be due to a bug in docker-for-osx. I'll try it with VirtualBox or something instead.

What do you use for the windows build? Do you build it on a windows box or from linux? Which VM setup do you use for gitian? I'd like to figure out how to get it working the same way you already do it...

dgenr8:
@jeffgran I'll post my gitian VM and send you a link.  Hopefully it's useful as a reference, but you don't want to copy from it.  The idea is to re-create your own independent build environment so you can verify binaries we produce.

jeffgran:
Happy to report that I was able to successfully build the osx app (against the HEAD of this branch -- `1f27f7166acca0e6c151b85b852d733bbc299eea`) with the existing gitian config just now. Output is below:

```
$ ./bin/gbuild --commit bitcoin=osx-app-build-updates ../bitcoinxt/contrib/gitian-descriptors/gitian-osx.yml 

--- Building for trusty amd64 ---
Stopping target if it is up
Making a new image copy
Starting target
Checking if target is up.
Preparing build environment
Updating apt-get repository (log in var/install.log)
Installing additional packages (log in var/install.log)
Upgrading system, may take a while
Creating package manifest
Creating build script (var/build-script)
Running build script (log in var/build.log)
Grabbing results from target
Generating report
6f9da94e67c4c1c9d1b89c2a0bb1d19aea5a1e6319109dac8becacfd536d0c73  bitcoin-xt-0.11.0-J-osx-unsigned.dmg
a3be5579abe398c475006dabbbcaddda8c19461ab9c6bfdaf3aa06e94d6f4594  bitcoin-xt-0.11.0-J-osx-unsigned.tar.gz
038480ac775711f8b3e1d807cb1485c12567736c96eb1bc17d6b0b03e698c619  bitcoin-xt-0.11.0-J-osx64.tar.gz
6e815177dc737dc73d23f92f5316d824243501e460d20469a009be598aa5c1fd  src/bitcoin-xt-0.11.0-J.tar.gz
e5143751aa346171415ee5452d569a64032b5c05cefaacb3dce23a5259ddeae2  bitcoin-osx-0.11-res.yml
Done.
```

I installed the app from the generated .dmg, double checked the updated `CFBundleIdentifier` in the `Info.plist`, and ran the app. It all seems to work fine. I think we can safely merge this. I believe @dgenr8 is attempting a build too. I think if it's against this same commit, we should see identical hashes of the built binaries?

dgenr8:
I also successfully built using the OSX 10.11 tarball you provided.  I built the I release so hashes don't match.

Successfully tested install and run.  The one problem I saw was the default data directory, which is still "Bitcoin".  This should be BitcoinXT since we're trying to avoid collisions.

dgenr8:
Merging without directory change, thanks again @jeffgran .

jeffgran:
Sorry for dragging my feet on this thanks for getting it in @dgenr8 !

PR #445, opened 2018-06-11
Replace stealth-mode with custom user agent

### LINKED ISSUE #443: Stealth mode broken on BCH
The stealth mode (`-stealth-mode=1`) which was introduced during the XT targeted DDoS attacks in 2015 makes XT masquerade as an older version of Bitcoin Core. 

This obviously won't fool anyone on the Bitcoin Cash network. Suggest we either make XT masquerade as a different BCH client or completely remove this feature.

dagurval:
This removes stealth mode and replaces it with the option to set a custom user agent.

If we want to enable full stealth mode again, we can implement the options to disable specific features to more closely mimic a client (for example `-useragent="/BUCash:1.3.0.1/" -disablecompactblocks -disablebip64` etc.).

Closes #443

### DIFF
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -129,8 +129,8 @@ std::string FormatFullVersion()
     return CLIENT_BUILD;
 }
 
-/** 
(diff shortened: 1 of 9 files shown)


dgenr8:
utACK

PR #442, opened 2018-06-03
Parallel initialization of UTXO commitment

dagurval:
Use multiple threads to calculate UTXO commitment from a coins view. On my system it takes 65 seconds to calculate the utxo commitment for mainnet using `-par=16` (16 threads).

(This is on top of #439 and #425)

Suggest cherry-picking this into #425 @tomasvdw 
```
2018-06-03 14:12:30 Generating UTXO commitment; progress 0
2018-06-03 14:12:32 Generating UTXO commitment; progress 2
2018-06-03 14:12:33 Generating UTXO commitment; progress 4
2018-06-03 14:12:35 Generating UTXO commitment; progress 7
2018-06-03 14:12:36 Generating UTXO commitment; progress 9
2018-06-03 14:12:38 Generating UTXO commitment; progress 12
2018-06-03 14:12:40 Generating UTXO commitment; progress 14
2018-06-03 14:12:41 Generating UTXO commitment; progress 17
2018-06-03 14:12:43 Generating UTXO commitment; progress 19
2018-06-03 14:12:45 Generating UTXO commitment; progress 22
2018-06-03 14:12:46 Generating UTXO commitment; progress 24
2018-06-03 14:12:48 Generating UTXO commitment; progress 27
2018-06-03 14:12:50 Generating UTXO commitment; progress 29
2018-06-03 14:12:51 Generating UTXO commitment; progress 32
2018-06-03 14:12:53 Generating UTXO commitment; progress 34
2018-06-03 14:12:54 Generating UTXO commitment; progress 37
2018-06-03 14:12:56 Generating UTXO commitment; progress 39
2018-06-03 14:12:58 Generating UTXO commitment; progress 42
2018-06-03 14:12:59 Generating UTXO commitment; progress 44
2018-06-03 14:13:01 Generating UTXO commitment; progress 47
2018-06-03 14:13:02 Generating UTXO commitment; progress 49
2018-06-03 14:13:04 Generating UTXO commitment; progress 52
2018-06-03 14:13:06 Generating UTXO commitment; progress 54
2018-06-03 14:13:07 Generating UTXO commitment; progress 57
2018-06-03 14:13:09 Generating UTXO commitment; progress 59
2018-06-03 14:13:11 Generating UTXO commitment; progress 62
2018-06-03 14:13:12 Generating UTXO commitment; progress 64
2018-06-03 14:13:14 Generating UTXO commitment; progress 67
2018-06-03 14:13:16 Generating UTXO commitment; progress 69
[...]

PR #446, opened 2018-06-11
Default to virtual cores and set appropriate thread scheduler

dagurval:
This bumps number of scriptcheck threads to # virtual cores, without re-introducing the issue that using physical core count fixed (see https://github.com/bitcoin/bitcoin/pull/6361).

This should give nice speedup to #442, which reuses the `-par` parameter.

No merge conflicts.
https://github.com/bitcoin/bitcoin/pull/10271 - Use std::thread::hardware_concurrency, instead of Boost, to determine available cores
https://github.com/bitcoin/bitcoin/pull/12618 - Set SCHED_BATCH priority on the loadblk thread.
https://github.com/bitcoin/bitcoin/pull/12923 - util: Pass pthread_self() to pthread_setschedparam instead of 0

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -591,6 +591,7 @@ void CleanupBlockRevFiles()
 void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
 {
     RenameThread("bitcoin-loadblk");
+    ScheduleBatchPriority();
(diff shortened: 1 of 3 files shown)


dagurval:
Thanks for the review @sickpig !

PR #447, opened 2018-06-12
Easy to use gitian building script

dagurval:
https://github.com/bitcoin/bitcoin/pull/8566 - Easy to use gitian building script
https://github.com/bitcoin/bitcoin/pull/11391 - Remove lxcbr0 lines from gitian-build.sh
f6b3382 (squashed into next commit), b9e79ab and 564a172 from https://github.com/bitcoin/bitcoin/pull/10773/
https://github.com/bitcoin/bitcoin/pull/12394 - gitian-builder.sh: fix --setup doc, since lxc is default
https://github.com/bitcoin/bitcoin/pull/12527 - gitian-build.sh: fix signProg being recognized as two parameters
https://github.com/bitcoin/bitcoin/pull/13320 - Ensure gitian-build.sh uses bash

Tested by building Windows and Linux versions for 0.11I and the sha256sums match.

Documentation for using this is located here: https://github.com/bitcoin-core/docs/blob/master/gitian-building.md

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -40,12 +40,15 @@ for Debian-based Linux systems. If you compile bitcoind/bitcoin-qt yourself, the
 Gavin's notes on getting gitian builds up and running using KVM.
 
 ### [Gitian-downloader](/contrib/gitian-downloader)
-Various PGP files of core developers. 
+Various PGP files of core developers.
 
 ### [MacDeploy](/contrib/macdeploy) ###
-Scripts and notes for Mac builds. 
+Scripts and notes for Mac builds.
 
-Test and Verify Tools 
+### [Gitian-build](/contrib/gitian-build.sh) ###
+Script for running full gitian builds.
+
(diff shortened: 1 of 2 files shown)


dgenr8:
@xt-bot retest this

dagurval:
Did you use your existing Gitian VM? If so, is it the same version of debian as [in the documentation?](https://github.com/bitcoin-core/docs/blob/master/gitian-building/gitian-building-create-vm-debian.md)

dgenr8:
@dagurval Thanks, turns out I was expecting it to do much more than it does.  I built all three platforms fine.  Should we turn on OSX by default?

ACK

dagurval:
> Should we turn on OSX by default?

I disabled it because we haven't released OSX binaries for some time. I'm all for enabling it if we have the resources to maintain minimal support for OSX (maybe @jeffgran is interested?).

PR #448, opened 2018-06-12
Move utxocommit from consensus to common

dagurval:
Fixes linker error on OSX.

The file utxocommit.cpp uses functionality that is not part of libbitcoinconsensus, including "compressor.cpp"


dgenr8:
@tomasvdw Ok with this change? It seems you would want this in the other repos also.

PR #450, opened 2018-06-12
Tuple syntax fix

dgenr8:
Remove C++17 syntax that sneaked in.


dagurval:
utACK, thanks for fixing

PR #457, opened 2018-06-19
Remove estimate priority

dagurval:
Estimate priority and code using it is dead code. We don't support transaction priority when generating block template anymore and there are as far as I know also no miners running node software that supports it on the network.

This includes https://github.com/bitcoin/bitcoin/pull/7730 (Remove priority estimation). Some patching required because we never added the "smart priority estimate" features.

Miner version was bumped because the estimate database file needed new versioning.


dagurval:
Pushed a commit to this PR that removes _priority delta_ from mempool.

dagurval:
Updated with suggested changes

PR #458, opened 2018-06-19
Move transaction priority functions into policy

dagurval:
Transaction priority is rules for which low fee transactions we're willing to include in the blocks we mine. It's policy.

On top of #457


dagurval:
rebased

dgenr8:
Ok, core has removed priority completely so we can refactor this without unnecessary divergence.

PR #459, opened 2018-06-20
 Introduce MempoolFeeModifer

dagurval:
This moves the fee delta functionality to a helper class, with it's own
mutex. Less responsibility for CTxMempool class

On top of #457 #458

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -181,6 +181,7 @@ BITCOIN_CORE_H = \
   util.h \
   utildebug.h \
   utilfork.h \
+  utilhash.h \
   utilmoneystr.h \
   utilprocessmsg.h \
   utiltime.h \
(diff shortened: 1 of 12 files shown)


dgenr8:
rebased-as-dagurval

dagurval:
109f4c6 is used in 9b58365

The thought was that it is easy to tell if you're working on _hash_ or _id_ while working on specific code segments. It's also easy to revert if we go for a malfix solution that doesn't need to differentiate.

It's difficult/boring and error prone to go through the whole code base and evaluate every use in one go later.

So for the sake of doing less work later, start differentiating now.

dgenr8:
Currently "hash" and "id" are interchangeable. "txid" is used in many variable names particularly in tests that are not updated here.  If we want to do a refactor now to prepare for that potential hard fork, I suggest changing all names to "hash".  Actually, no I don't because yuck, a useless refactor.

MempoolFeeModifier looks good.  @dagurval If the partial refactor is really important to you I will ACK it.

dagurval:
I've removed 109f4c6 👍

dgenr8:
ACK, thanks, I'm much happier with this

PR #461, opened 2018-06-25
Refactor tx fee/priority evaluation out of AcceptToMemoryPool

dagurval:
This split out transaction fee/priority evaluation out of `AcceptToMemoryPool` and into a class `FeeEvaluator`. This class is given unit test coverage, and the tests clearly define transaction acceptance policy.

The somewhat broken `-relaypriority` parameter is replaced with a more descriptive `-allowfreetx` parameter.

QA tests that broke due to fee policy change are fixed.

On top of #459

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -982,21 +982,24 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
 
         CTxMemPoolEntry entry(tx, nFees, GetTime(), chainActive.Height(), pool.HasNoInputsOf(tx), fSpendsCoinbase, lp, nSigOps);
 
-        // Don't accept it if it can't get into a block
+        FeeEvaluator feeEval(Opt().AllowFreeTx(), mempool.GetFeeModifier(),
+                             ::minRelayTxFee);
+        FeeEvaluator::FeeState feestate = feeEval.HasSufficientFee(view, entry,
+                                                                   chainActive.Height());
         if (!fLimitFree) {
-            FeeEvaluator feeEval(Opt().AllowFreeTx(), mempool.GetFeeModifier(), ::minRelayTxFee);
-            auto res = feeEval.HasSufficientFee(view, entry, chainActive.Height());
-            if (res != FeeEvaluator::FEE_OK) {
-                std::string err = FeeEvaluator::ToString(res);
+            // Don't accept it if it can't get into a block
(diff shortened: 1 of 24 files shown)


dgenr8:
Rebased-as-@dagurval

src/main.cpp
```diff
@@ -1006,39 +981,18 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
+            FeeEvaluator feeEval(Opt().AllowFreeTx(), mempool.GetFeeModifier(), ::minRelayTxFee);
+            auto res = feeEval.HasSufficientFee(view, entry, chainActive.Height());
+            if (res != FeeEvaluator::FEE_OK) {
+                std::string err = FeeEvaluator::ToString(res);
+                return state.DoS(0, false, REJECT_INSUFFICIENTFEE, err);
+            }
         }
 
+        unsigned int nSize = entry.GetTxSize();
         if (fRejectAbsurdFee && nFees > ::minRelayTxFee.GetFee(nSize) * 10000)
```
dgenr8: Should the absurd-fee logic also be part of FeeEvaluator?

dagurval: Done in a new commit

src/miner.cpp
```diff
@@ -99,6 +99,9 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
     // Limit to between 1K and (hard limit - 1K) for sanity:
     nBlockMaxSize = std::max((uint64_t)1000, std::min((hardLimit - 1000),  nBlockMaxSize));
 
+    // For compatibility with bip68-sequence test, set flag to not mine txs with negative fee delta.
+    const bool fSkipNegativeDelta = GetBoolArg("-bip68hack", false);
```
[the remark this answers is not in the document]
dagurval: > It's possible a miner may want to relay certain transactions but not mine them himself. Any interest in naming this -minenegativedelta (even defaulting to false), ...

I'd like for a miner to present me with a use case where this is desirable, then I'd be happy to rename it. For now, I think it's better to keep the current name to emphasize it's a feature we want to remove.

> ... or removing the option completely, since our usage of fee delta is already unique to XT?

I'd like that, but the bip68 rpc test needs to change then.

dgenr8:
@dagurval Curious, do you see more of ATMP moving into mempoolacceptor?

dagurval:
@dgenr8 With time, I hope to see all of it moved and with tests :-)

PR #463, opened 2018-06-26
Disallow messages before receiving VERACK

dagurval:
* fd13eca from https://github.com/bitcoin/bitcoin/pull/9674
* https://github.com/bitcoin/bitcoin/pull/9715 - Disconnect peers which we do not receive VERACKs from within 60 sec
* https://github.com/bitcoin/bitcoin/pull/9720 - net: fix banning and disallow sending messages before receiving verack
    * c45b9f - I already wrote the same function as SendRejectsAndCheckIfBanned. Moved it to utilprocessmsg instead of moving it up in main.
    * skipped 5b5e4f8 and 8650bbb, we already got them in 9715

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5676,8 +5676,10 @@ bool ProcessMessages(CNode* pfrom, CConnman* connman, std::atomic<bool>& interru
             PrintExceptionContinue(NULL, "ProcessMessages()");
         }
 
-        if (!fRet)
+        if (!fRet) {
             LogPrint(Log::NET, "%s(%s, %u bytes) FAILED peer=%d\n", __func__, SanitizeString(strCommand), nMessageSize, pfrom->id);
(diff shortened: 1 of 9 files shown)


src/main.cpp
```diff
@@ -4756,9 +4756,12 @@ bool ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
         }
         pfrom->fSuccessfullyConnected = true;
     }
-
-
-    else if (strCommand == "addr")
+    else if (!pfrom->fSuccessfullyConnected)
+    {
+        Misbehaving(pfrom->GetId(), 1, "must have a verack message before anything else");
```
dgenr8: ACK removal of LOCK(cs_main) in source commit.

dgenr8:
ACK

PR #469, opened 2018-07-06
Add experimental support for LMDB

dagurval:
On top of #468 

LMDB is a high-performance embedded transactional database in the form of a key-value store. It's minimal and focused on modern hardware, 64bit CPU & SSD.

This PR makes use of lmdb optional, by passing the parameter `-db=lmdb`.

Performance results on my development machine:

**Rotating disk:**
$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**lmdb** -reindex-chainstate
real    **63m39**,459s
user    15m29,719s
sys     1m2,658s

$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**leveldb** -reindex-chainstate
real    **24m40**,074s
user    23m9,927s
sys     2m27,688s

**SSD:**

Testes with a 6 year old SSD disks.

$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**lmdb** -reindex-chainstate; 
real    **17m56**,519s
user    14m21,408s
sys     1m52,948s

$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**leveldb** -reindex-chainstate
real    **25m36**,895s
user    23m43,988s
sys     4m2,225s

**SSD without forced OS flushing**

I also added the option `-dbsafemode`, which, when disabled removes forced disk flush for every transaction commit. This gives performance boost, at the cost of higher risk of db corruption in case of abnormal program termination.

$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**lmdb** -reindex-chainstate -dbsafemode=0
real    **16m49**,922s
user    14m21,606s
sys     1m36,634s

**Smaller dbcache**
Because of the memory mapping mechanism, the performance is less dependent on the size of dbcache set. I suspect the performance degradation here is due to CPU being busy with serialization and not I/O blockage.

$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**lmdb** -reindex-chainstate -dbsafemode=0 -dbcache=100

real    **17m28**,172s
user    14m13,753s
sys     2m5,633s

$ time ./src/bitcoind -stopatheight=300000 -testnet=0 -db=**leveldb** -reindex-chainstate -dbsafemode=0 -dbcache=100

real    **33m30**,324s
user    32m24,312s
sys     7m36,293s

**Future work:**
- Auto-grow max database size. LMDB requires that you set max database size up front. This can be changed under certain conditions during runtime.
- Warn if used with rotating disks.
- Implement utxo size estimation
- Investigate more optimal use of LMDB properties 
    - Can we use the lock-free parallel read support?
    - Can we remove dbcache altogether
- [qa] Fix qa tests with leveldb assumptions (bip100-sizelimit, blockchain, walletbackup)
- [qa] Remove cache if it's for the wrong db backend

Except for these, all qa tests pass:
* bip100-sizelimit.py - bypasses framework for bitcoin.conf, thus uses leveldb
* blockchain.py - tests for size, not implemented for lmdb
* walletbackup - assumes blockstate folder

laanwj has an experimental branch replacing leveldb with lmdb in 2016, here is their notes: https://github.com/laanwj/bitcoin/blob/2016_04_mdb/README.md. This work is in part based on laanwj's implementation. LMDB is also used by Monero.


src/dbwrapper.h
```diff
@@ -25,190 +23,145 @@ class dbwrapper_error : public std::runtime_error
...
+};
 
+class ClearStreamRAII {
+public:
+    ClearStreamRAII(CDataStream& s1) : s1(&s1), s2(nullptr) { }
+    ClearStreamRAII(CDataStream& s1, CDataStream& s2) : s1(&s1), s2(&s2) { }
+    ~ClearStreamRAII() {
+        s1->clear();
+        if (s2 != nullptr) {
+            s2->clear();
```
awemany: Hmm, wouldn't it be straightforward to use two `ClearStreamRAII` instead, one for each stream, and then simplify the implementation of this class?

(I guess I should (and will) stick with BU but @sickpig pointed me to this interesting PR - and this bit just strikes me as odd :D)

dagurval: That makes sense. No need to stick to BU 😁

dagurval:
Added dynamic DB mapsize adjusting and utxo size estimation.

dgenr8:
@kyuupichan Can you explain a little about the scenario where you found LMDB did not perform well? Do you recall whether you were using rotating disk or SSD?

kyuupichan:
@dgenr8 I was trying to use LMDB for (address) history storage in ElectrumX.  We had it for a while but couldn't make it perform well; once we got to height about 250k it was crawling compared to leveldb.  The access pattern of history is purely appending new key-values; no lookup at all is done.  i.e. it's write-only.  I used a spinning disk.  My results were confirmed by @bauerj; maybe he can chime in on his configuration.  

See also https://github.com/kyuupichan/electrumx/commit/c958b3af494094aa9b99e7c153344444c0b51cb2.

It is possible we were not using LMDB in the right way or optimally, but we did try.

dagurval:
Added disk space check to `DBResizeManager`. Added unit tests for `DBResizeManager`.

PR #470, opened 2018-07-20
Fix read of timestamp after free

dagurval:
Timestamp that's used in debug output was read after free. Detected by valgrind.


dgenr8:
utACK

PR #474, opened 2018-08-21
[fork] Helper functions and mempool triggers

dagurval:
Helper functions to check for upgrade activation. Triggers to clean mempool on activation/deactivation. 

Also includes a _revert me_ commit, that when reverted, enables the upgrade by default.

### DIFF
--- a/src/test/utilfork_tests.cpp
+++ b/src/test/utilfork_tests.cpp
@@ -239,5 +239,36 @@ BOOST_AUTO_TEST_CASE(forkmempoolclearer_thirdhf) {
     BOOST_CHECK_EQUAL(1, mempool.clearCalls);
 }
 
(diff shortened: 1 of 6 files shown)


src/test/utilfork_tests.cpp
```diff
@@ -240,6 +240,10 @@ BOOST_AUTO_TEST_CASE(forkmempoolclearer_thirdhf) {
 }
 
 BOOST_AUTO_TEST_CASE(forkmempoolclearer_fourthhf) {
+    auto arg = new DummyArgGetter;
+    auto argraii = SetDummyArgGetter(std::unique_ptr<ArgGetter>(arg));
+    arg->Set("-fourthhftime", 1526400000); // TODO: REVERT ME!
```
magnusga: Shouldn't the timestamp be 1542300000?

dagurval:
Fixed timestamps

PR #478, opened 2018-09-04
Add operator> to base_blob and use proper endianess

dagurval:
* https://github.com/bitcoin/bitcoin/pull/7712 - Improve COutPoint less operator
* https://reviews.bitcoinabc.org/D1526 - Add support for both way comparison for uint256/uint160
    * Random code formatting changes removed
* https://reviews.bitcoinabc.org/D1527 - Modify uint256's comparison to use proper endianess
    * Changed to const reference for `a` and `b` in `base_blob::Compare`

As far as I can tell, nothing depended on the existing behavior of `base_blob::Compare`, only that there was an order. This is a prerequisite for CTOR.

### DIFF
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -36,7 +36,8 @@ public:
 
     friend bool operator<(const COutPoint& a, const COutPoint& b)
     {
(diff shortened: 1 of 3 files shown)


dgenr8:
utACK

PR #479, opened 2018-09-08
Throw detailed error from DecodeHexTx

dagurval:
This improves the output from the RPC interface.

```
decoderawtransaction 0100000001db4f85b6e2b9f6378fd9f52d80e17c63153D61e1cbc4d46632bf2aba3190d5b000000000000000000001000000000000000001510000000000

TX decode failed: Excess data in hex string. After parsing transaction, 1 bytes remained. (code -22)
```


dgenr8:
utACK. This is way cooler than what the other guys have :sunglasses:

PR #482, opened 2018-09-12
[test] Add helper function BuildDummyBlockIndex

dagurval:
Make code for chaining a dummy index reusable with new function `BuildDummyBlockIndex`.

I plan to use this with #477


dgenr8:
utACK, more niceness from the lambda lover

dagurval:
Better to fancy lambdas than lambos in this bear market :grin:

PR #483, opened 2018-09-14
[fork] Remove ordering requirement in RollforwardBlock

dagurval:
This makes RollforwardBlock support any transaction order.

It's a prerequisite for supporting AOR (Any-order-rule) and CTOR.

https://reviews.bitcoinabc.org/D1479

Changes from original commit:
* Removed optimization in `UpdateCoins(tx, inputs, nHeight)` - that optimization removes the need to copy txundo, but duplicates code.
* Did not add header declaration for SpendCoins, but made it static instead.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1392,20 +1392,27 @@ void static InvalidBlockFound(CBlockIndex *pindex, const CValidationState &state
     }
 }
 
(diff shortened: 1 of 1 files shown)


dagurval:
@xt-bot retest this

PR #484, opened 2018-09-15
Update best known in nodestate and add tests

dagurval:
Move ProcessBlockAvailability into CNodeState as UpdateBestFromLast, add
tests and remove duplicate code in UpdateBlockAvailability


dagurval:
`HasLessWork` should be `HasLessOrEqualWork`. If not, we won't download other block candidates with equal work. Fixed and updated tests.

PR #485, opened 2018-09-17
[fork] Introduce Any-Order-Rule

dagurval:
* Make a copy of PrecomputedTransactionData in the CScriptCheck - https://reviews.bitcoinabc.org/D297
    * Did not move the header declaration of PrecomputedTransactionData
    * Did not add a copy constructor (we'll rely on implicitly-declared)

* Process transactions in two loops in ConnectBlock instead of one - https://reviews.bitcoinabc.org/D1484
    * Needed some patching. ABC has refactored the sigop counting.

* Remove topological ordering constraint from block starting Nov, 15 2018 - https://reviews.bitcoinabc.org/D1487
    * Test needed some patching due to test framework differences

* Update ApplyBlockUndo to be able to undo out of order blocks - https://reviews.bitcoinabc.org/D1528
    * Compiler started to complain about returning `int`, so I converted DisconnectResult into an `enum class`. The `class` attribute allows it to be forward declared in the tests.
    * Some minor patching due to ABC refactors

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -127,6 +127,7 @@ testScripts = [
     'abc-monolith-activation.py',
     'bip64.py',
     'p2p-leaktests.py',
+    'abc-transaction-ordering.py',
 ]
 
 testScriptsExt = [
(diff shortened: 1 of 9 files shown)


dagurval:
> I suggest we save up all the ABC fork changes and merge together.

I think that's the wrong approach for two reasons:
1) More changes at the same time, increases the chance of error and makes debugging more difficult.
2) Makes it more difficult for me to implement the final changes, as I'll have to do it on top of multiple PR's

dgenr8:
Rebased

PR #487, opened 2018-09-21
[fork] Impose a minimum transaction size of 100 bytes

dagurval:
https://reviews.bitcoinabc.org/D144
https://reviews.bitcoinabc.org/D1611

### DIFF
--- a/qa/rpc-tests/bip68-sequence.py
+++ b/qa/rpc-tests/bip68-sequence.py
@@ -12,6 +12,7 @@ from test_framework.util import *
 from test_framework.script import *
 from test_framework.mininode import *
 from test_framework.blocktools import *
+from test_framework.txtools import bloat_tx
 
 SEQUENCE_LOCKTIME_DISABLE_FLAG = (1<<31)
 SEQUENCE_LOCKTIME_TYPE_FLAG = (1<<22) # this means use time (0 means height)
@@ -86,6 +87,7 @@ class BIP68Test(BitcoinTestFramework):
         sequence_value = sequence_value & 0x7fffffff
         tx2.vin = [CTxIn(COutPoint(tx1_id, 0), nSequence=sequence_value)]
         tx2.vout = [CTxOut(int(value-self.relayfee*COIN), CScript([b'a']))]
+        bloat_tx(tx2)
         tx2.rehash()
 
         try:
@@ -235,7 +237,8 @@ class BIP68Test(BitcoinTestFramework):
             tx.nVersion = 2
             tx.vin = [CTxIn(COutPoint(orig_tx.sha256, 0), nSequence=sequence_value)]
             tx.vout = [CTxOut(int(orig_tx.vout[0].nValue - relayfee*COIN), CScript([b'a']))]
-            tx.rehash()
+            bloat_tx(tx)
+            tx.calc_sha256()
 
             try:
                 node.sendrawtransaction(ToHex(tx))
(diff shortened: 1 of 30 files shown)


dagurval:
@sickpig @ptschip @schancel You may find commit `[qa] Introduce bloat_tx and bloat_raw_tx` useful.

Also I've extended the ABC test to test ATMP in `[qa] Add ATMP test for undersize transaction`

dagurval:
@xt-bot retest this

schancel:
@dagurval That's basically a helper pulled out of the other test? txtools module is probably a good thing to start.  I've had to do a lot of fixing up on the tests that generate transactions.

dagurval:
@schancel Sorry, I should explain. It looks to me that ABC has not attempted to activate the November upgrade on the QA test framework.

In our case when I did this, several of the qa tests started to fail due to tx-undersize. I think you will have this issue too.

I just wanted to point out here is a solution if you want it.

qa/rpc-tests/bip68-sequence.py
```diff
@@ -235,7 +237,8 @@ def test_nonzero_locks(orig_tx, node, relayfee, use_height_lock):
             tx.nVersion = 2
             tx.vin = [CTxIn(COutPoint(orig_tx.sha256, 0), nSequence=sequence_value)]
             tx.vout = [CTxOut(int(orig_tx.vout[0].nValue - relayfee*COIN), CScript([b'a']))]
-            tx.rehash()
+            bloat_tx(tx)
+            tx.calc_sha256()
```
jasonbcox: Why switch from tx.rehash() to tx.calc_sha256()?  They're effectively equal here (and the other places where this change was made)

qa/rpc-tests/p2p-fullblocktest.py
```diff
@@ -629,6 +632,7 @@ def update_block(block_number, new_transactions):
             scriptSig = CScript([sig, redeem_script])
 
             tx.vin[1].scriptSig = scriptSig
+            bloat_tx(tx)
```
jasonbcox: Wouldn't calling bloat_tx() in tx.rehash() would make this change more pervasive and less error prone? (not to mention the nice side effect of fewer line-changes :) )

qa/rpc-tests/test_framework/blocktools.py
```diff
@@ -74,6 +73,8 @@ def create_transaction(prevtx, n, sig, value, scriptPubKey=CScript()):
     tx.vin.append(CTxIn(COutPoint(prevtx.sha256, n), sig, 0xffffffff))
     tx.vout.append(CTxOut(value, scriptPubKey))
     tx.calc_sha256()
```
jasonbcox: This should come after the padding step.

qa/rpc-tests/p2p-fullblocktest.py
jasonbcox: Then again, having that side effect where it's not expected might make debugging difficult...

qa/rpc-tests/bip68-sequence.py
dagurval: It's just a small optimization, `calc_sha256` only calculates the hash if it has not been done before. I know bloat_tx does a rehash, but I don't want to depend on this side effect.

qa/rpc-tests/test_framework/blocktools.py
dagurval: Thanks :+1:

qa/rpc-tests/p2p-fullblocktest.py
jasonbcox: Ya, realized my mistake when I started poking around at the code some more myself.

qa/rpc-tests/test_framework/blocktools.py
dgenr8: Needs fix

dgenr8:
Rebased on top of #485.

Issue #494, opened 2018-10-06
Support block files > 2GiB on windows


dgenr8:
Factor out reliance on fseek(... long int ...) and anything else preventing use of > 2GiB files.

dgenr8:
#495 is a workaround that makes us not try to generate files > 2GiB.

PR #495, opened 2018-10-06
Keep block files under 2GB thanks to windows fseek

dgenr8:
Until we can test a fix for issue #494, set MinBlockFileBlocks to 64.  This needs to be revisited when the max block size is increased beyond 32MB.


dgenr8:
The setting of 64 gets us back to Sept 1, so about one rollover every 45 days.  This was during the stress test. Going back another 45 days only adds about 32 more max-sized blocks.

PR #496, opened 2018-10-19
Encapsulate relay cache + new tests

dagurval:


sickpig:
Ok this was quick! Keep up the good work @dagur! :smile:

src/main.cpp
```diff
@@ -4280,10 +4281,9 @@ void static ProcessGetData(CNode* pfrom, CConnman* connman, std::atomic<bool>& i
                 // Send stream from relay memory
                 bool pushed = false;
                 {
-                    LOCK(cs_mapRelay);
-                    map<CInv, CDataStream>::iterator mi = mapRelay.find(inv);
-                    if (mi != mapRelay.end()) {
-                        connman->PushMessage(pfrom, NetMsg(pfrom, inv.GetCommand(), (*mi).second));
+                    CTransaction tx = RelayCache::Instance().FindTx(inv.hash);
+                    if (!tx.IsNull()) {
+                        connman->PushMessage(pfrom, NetMsg(pfrom, inv.GetCommand(), tx));
```
dgenr8: Wasn't avoiding a re-serialization here the purpose of caching the data stream?  Was this not worth it?

src/relaycache.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return cache;
+}
+
+void RelayCache::ExpireOld() {
+    std::unique_lock<std::mutex> lock(cs);
+    if (expiration.empty()) {
+        return;
+    }
+    int64_t time = GetTime();
+    while (expiration.front().first <= time) {
```
dagurval: should be `expiration.empty() && expiration.front() ...` here

src/main.cpp
dagurval: It adds some complexity, but it may be worth it. I'd have to write a benchmark test to know for sure.

Core also changed to using CTransaction, but for segwit reasons: https://github.com/bitcoin/bitcoin/pull/7877

Relevant comment from sdaftuar:

> In case anyone else has the same thought: I was wondering what the memory usage ratio was between a CTransaction and its serialization; it seems to be about 1.4x as big [..segwit stuff..] (Also serialization seems to be fast enough that it's not worth worrying about doing it for each peer.)

dgenr8: For that matter, the original received serialization could be cached when the CTransaction is created.  A drawback would be less visibility of deser/reser bugs, and it's a topic for some other time/place.

src/test/relaycache_tests.cpp
```diff
@@ -0,0 +1,77 @@
...
+    cache.Insert(tx1);
+    cache.time = 101;
+    cache.Insert(tx2);
+
+    BOOST_CHECK(cache.FindTx(tx1.GetHash()) == tx1);
+    BOOST_CHECK(cache.FindTx(tx2.GetHash()) == tx2);
+
+    // Expire tx1
+    cache.time = 100 + RELAY_CACHE_TIMEOUT;
+    BOOST_CHECK(cache.FindTx(tx1.GetHash()).IsNull());
```
dgenr8: Since ExpireOld() is not called, expiration is not being tested.  IsNull() passes because setting the version is not sufficient to make it not pass.

PR #499, opened 2018-10-26
Convert signature hash type to enum class

dagurval:
Gives signature hash types a typename, thus adding type safety.

This is a less intrusive alternative to the `SigHashType` enum wrapper class introduced by ABC. Also this is an alternative to BU's approach to keep both the enum and introduce ABC's class, as done in their DSV PR https://github.com/BitcoinUnlimited/BitcoinUnlimited/pull/1390

This change makes it easier to port signature encoding tests and DSV.

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -17,6 +17,7 @@
 #include "rpc/server.h"
 #include "script/script.h"
 #include "script/script_error.h"
+#include "script/sighashtype.h"
 #include "script/sign.h"
 #include "script/standard.h"
 #include "uint256.h"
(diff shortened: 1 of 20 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -755,30 +756,18 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
-            (std::string("SINGLE|ANYONECANPAY"), int(SIGHASH_SINGLE | SIGHASH_ANYONECANPAY))
-            (std::string("SINGLE|FORKID"), int(SIGHASH_SINGLE | SIGHASH_FORKID))
-            (std::string("SINGLE|FORKID|ANYONECANPAY"), int(SIGHASH_SINGLE | SIGHASH_FORKID | SIGHASH_ANYONECANPAY));
         string strHashType = request.params[3].get_str();
-        if (mapSigHashValues.count(strHashType))
-            nHashType = mapSigHashValues[strHashType];
-        else
-            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid sighash param");
+        try {
+            nHashType = FromStr(strHashType);
```
dgenr8: Do the other implementations have a dependency on flag ordering as this does?

dagurval: Looks like BU does not have a dependency on ordering, but will also not tell you when you mess up (passing ALL|NONE magically converts into SINGLE).


 https://github.com/BitcoinUnlimited/BitcoinUnlimited/blob/release/src/rpc/rawtransaction.cpp#L820

dgenr8:
@dagur Could you state something about how this refactor avoids a problem similar to the one that [affected ABC](https://www.bitcoinabc.org/2018-05-07-incident-report/)?

dagurval:
This change is less bug prone compared to ABC's approach as it does not change the interface.

This line:
```
flags = SIGHASH_ALL | SIGHASH_FORKID;
```
is the same with this approach:
```
flags = SigHashType::ALL | SigHashType::FORKID;
```
while with ABC it's:
```
flags = SigHashType().withForkId();
```

Another example:
```
bool fHasSingle = (nHashType & ~(SIGHASH_ANYONECANPAY | SIGHASH_FORKID)) == SIGHASH_SINGLE
```
this approach:
```
bool fHasSingle = (nHashType & ~(SigHashType::ANYONECANPAY | SigHashType::FORKID)) == SigHashType::SINGLE
```
ABC is completely different:
```
bool fHasSingle = nHashType.withAnyoneCanPay(false).withForkId(false).getRawSigHashType() == SIGHASH_SINGLE;
```

Errors in updating the use of sighash flags are harder to make and easier to spot.

src/rpc/rawtransaction.cpp
sickpig: Ouch this is not good for BU. Thanks for the report @dagurval!

dagurval: @sickpig I looked again and missed a detail, you have a `=` and not `|=` for the base flags.

This is even worse, because flag ordering is required but not enforced. As it stands `FORKID|SINGLE` results in `SINGLE`. `FORKID|ANYONECANPAY|SINGLE|NONE` results in `NONE`. `NOFORKID|FORKID` results in a sighashtype with FORKID, but signs it the wrong way.

PR #516, opened 2018-11-30
Introduce BlockBuilder interface

dagurval:
The interface replaces use of CBlockTemplate. The interface allows for building different representations of a new block without going through CBlock.

Before, transactions in `getblocktemplate` would pass two steps:
>    Mempool -> (copy to) CBlock -> (copy to) JSON

Now it is:
>    Mempool -> (copy to) JSON

For now, for sanity check, the JSON is converted to CBlock for the purpose of validations. This should be skipped in the future.

Also fixes a bug where sigop and txfees are counted wrong in 'getblocktemplate' after CTOR.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -14,6 +14,9 @@
 #include "hash.h"
 #include "main.h"
 #include "maxblocksize.h"
+#include "miner/blockbuilder.h"
+#include "miner/serializableblockbuilder.h"
+#include "miner/utilminer.h"
 #include "net.h"
 #include "policy/policy.h"
 #include "pow.h"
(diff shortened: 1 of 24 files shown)


src/miner.cpp
```diff
@@ -206,68 +203,73 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
...
-            throw std::runtime_error(strprintf("%s: TestBlockValidity failed: %s", __func__, FormatStateMessage(state)));
+        block.SetHashPrevBlock(pindexPrev->GetBlockHash());
+        UpdateTime(block, Params().GetConsensus(), pindexPrev);
+        block.SetBits(GetNextWorkRequired(pindexPrev, block.GetTime(), Params().GetConsensus()));
+
+        if (block.NeedsGBTMetadata()) {
+            block.SetBlockMinTime(nMedianTimePast + 1);
+            block.SetBlockHeight(nHeight);
+            block.SetBlockSizeLimit(hardLimit);
+            block.SetBlockSigopLimit(MaxBlockSigops(nBlockSize));
```
dagurval: This is different from before, it used to be `MaxBlockSigops(hadLimit)`. I think it is more accurate to be on the safe side here.

src/rpc/mining.cpp
```diff
@@ -585,177 +575,22 @@ UniValue getblocktemplate(const JSONRPCRequest& request)
     }
 
     // Update block
```
dagurval: To simplify, we don't do template caching here anymore. I believe the cache invalidation was wrong anyway, as `setClientRules` was not respected. The end goal is *fast* `getblocktemplate` also.

dgenr8 (approved):
I really like gbtblockbuilder encapsulating so much of the special block construction logic.  Also the more consistent use of exceptions.

src/miner.cpp
dagurval: Since sigoplimit grows with block size, with `hardLimit` he'd have to stuff the block completely full to be sure he's not exceeding the limit when adding transactions from another source, right?

dgenr8: You're right of course.  How quickly I forgot the decision to validate using the actual block size.

DaunthE1 (commented):
iQIcBAEBCAAGBQJboV68AAoJEJDIAZ42wulkrD0P/iULbLc7SRAXPPaQDxRV+nXO

PR #519, opened 2019-05-17
Enforce per-tx limits on CDS/CDSV sigops

dgenr8:
In AcceptToMemoryPool, set the fork-dependent script verification flags prior to checking the transaction sigop limit.  The check was failing to count CDS/CDSV sigops as per 2018-11 fork spec.


dgenr8:
It was convenient to remove 2012 BIP16 activation logic is part of this change.

